PR #36, opened 2024-01-05
Feat: global nav

jared-dickman:
## Instructions

 ## Summary
- POC for introducing a Global Navigation component

 ## Testing Plan

 - Closes https://go.mparticle.com/work/UNI-68

## Figma Exporation
- https://www.figma.com/file/b9JXWRvOt5Ryw2Q27kyrdE/%5BFluffy%5D-Platform-re-skin?type=design&node-id=367-8139&mode=design&t=96ohvqgx84TpP5J0-0


src/components/navigation/GlobalNavigation/WorkspaceSelectorItems.d.ts
tibuurcio, line 1: As a typing suggestion, we could make TS know better what we are referring to through our types, something like this:

```ts
type OrgType = 'org' | 'account' | 'workspace'

export interface INavigationOrg extends IWorkspaceSelectorItem {
  accounts: INavigationAccount[]
  type: 'org'
}

export interface INavigationAccount extends IWorkspaceSelectorItem {
  workspaces: INavigationWorkspace[]
  type: 'account'
}

export interface INavigationWorkspace extends IWorkspaceSelectorItem {
  type: 'workspace'
  isSelected: boolean// todo: is this needed? 
  onClick: (info: MenuInfo) => void
}

export interface IWorkspaceSelectorMapping extends MenuItemType {
  type: OrgType
  className: string
  label: string
  id: string
  
  isSelected?: boolean// todo: is this needed? 
  accounts?: INavigationAccount[]
  workspaces?: INavigationWorkspace[]
}
```

image
image

src/components/navigation/GlobalNavigation/WorkspaceSelectorItems.d.ts
```diff
@@ -0,0 +1,31 @@
...
+  onClick: (info: MenuInfo) => void
+}
+
+export interface IWorkspaceSelectorMapping extends MenuItemType {
+  type: 'org' | 'account' | 'workspace'
+  className: string
+  label: string
+  id: string
+  
+  isSelected?: boolean// todo: is this needed?
```
tibuurcio, line 28: A selection example might not be the best example for this but I'm seeing a lot of booleans through our front-end codebase and whenever a see one like this I always like to follow an enumerable approach instead. For instance changing `isSelected` to something like `selectionStatus` and giving it values of `selected | deselected`.

This gives us much more room to add more options in the future. for instance, if we would to track hover status in the future we would be able to use the same variable and have 3 options instead of having to create another `isHovering` variable. Again not the best example but just as a general note of something for us to think about.

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -0,0 +1,87 @@
...
+  onClick?: (e: React.MouseEvent) => void
+}
+
+export interface IGlobalNavigationProps {
+  logo: IGlobalNavigationLogo
+  tools: IGlobalNavigationTool[]
+  management: IGlobalNavigationManagement[]
+
+  orgs?: INavigationOrg[]
+  createItems?: INavigationCreateProps['createItems']
```
tibuurcio, line 47: Since we will be only using it in Indicative I think we should consider moving the createItems prop to "userland" and have the GlobalNavigation component export a way to inject stuff in the place we want it. Something like a `centerContent` prop that accepts a ReactNode.

tibuurcio (commented):
Some suggestions, wonderful work so far!

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -0,0 +1,87 @@
...
+export interface IGlobalNavigationManagement extends IBaseGlobalNavigationItem {
+  type: 'menu' | 'link'
+  children?: MenuItem[]
+  onClick?: (e: React.MouseEvent) => void
+}
+
+export interface IGlobalNavigationProps {
+  logo: IGlobalNavigationLogo
+  tools: IGlobalNavigationTool[]
+  management: IGlobalNavigationManagement[]
```
tibuurcio, line 44: How do you feel about using header, content, footer and/or similar naming here? I think in the future we will need to configure those "sections" individually and having those named like this could help.

src/components/navigation/GlobalNavigation/NavigationList.tsx
```diff
@@ -0,0 +1,54 @@
...
+export interface INavigationListProps {
+  items: Array<IGlobalNavigationManagement | IGlobalNavigationTool>
+  onClick?: (item: Item) => void
+}
+
+export function NavigationList(props: INavigationListProps) {
+  return (
+    <Center vertical>
+      {props.items.map((item, i) => (
+        <>
```
tibuurcio, line 20: Think we need a key here

tibuurcio, line 20: Think we need a key here

src/components/navigation/GlobalNavigation/NavigationList.tsx
```diff
@@ -0,0 +1,54 @@
...
+    children = item.children.map((child, j) => ({
+      key: `${String(child.label)}${j}`,
+      ...child,
+    }))
+
+    children.unshift({ label: item.label, type: 'group', key: item.label + '_groupTitle' })
+  }
+
+  return {
+    icon: <NavigationIcon icon={item.icon} label={item.label} onClick={item.onClick} hideLabel={item.hideLabel} />,
```
tibuurcio, line 49: Shouldn't the onClick here be applied to the whole item container and not only the icon? I saw that antd's Menu component doesn't seem to support this very well and I'm not being able to click in nested children in mP. Something for us to investigate next week.

src/components/navigation/GlobalNavigation/NavigationItem.tsx
```diff
@@ -0,0 +1,35 @@
...
+  hideLabel?: boolean
+  items? // menu only
+  onClick?: (e) => void // link only
+}
+
+export function NavigationItem(props: INavigationIconProps) {
+  return (
+    <>
+      {props.type === 'link' && (
+        <>
```
tibuurcio, line 19: Do we need the extra fragment here?

design/LightTheme.ts
```diff
@@ -0,0 +1,392 @@
...
+    'fontSize-H2': 30,
+    'fontSize-H3': 24,
+    'fontSize-H4': 20,
+    'fontSize-H5': 16,
+    'fontSize-Icon': 12,
+    lineHeightSM: 1.6699999570846558,
+    lineHeightH1: 1.2100000381469727,
+    lineHeightH2: 1.2699999809265137,
+    lineHeightH3: 1.3300000429153442,
+    lineHeightH4: 1.399999976158142,
```
tibuurcio, line 145: Those big decimals always makes eslint complain 😞 
We might need to exclude the rule from this file since this is auto generated.

package.json
```diff
@@ -17,7 +17,7 @@
   "files": [
     "dist/aquarium.mjs",
     "dist/index.d.ts",
-    "dist/_variables.css"
+    "dist/style.css"
```
tibuurcio, line 20: Good call 👍🏼

src/components/general/Button/Button.stories.tsx
```diff
@@ -14,32 +14,32 @@ const meta: Meta<typeof Button> = {
     disabled: false,
     ghost: false,
     href: undefined,
-    htmlType: 'button',
+    htmlType: "button",
```
tibuurcio, line 17: Think this is incorrect

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { NavigationList } from 'src/components/navigation/GlobalNavigation/NavigationList'
+import { NavigationCreate } from 'src/components/navigation/GlobalNavigation/NavigationCreate'
+import { WorkspaceSelector } from 'src/components/navigation/GlobalNavigation/WorkspaceSelector'
+import { type INavigationOrg } from 'src/components/navigation/GlobalNavigation/WorkspaceSelectorItems'
+import MpLogo from 'src/assets/mpLogo.svg?react'
+
+export interface IBaseGlobalNavigationItem {
+  label: string
+  icon: ReactNode
+  hideLabel?: boolean
```
tibuurcio, line 20: Isn't it better to make `label` optional instead of having a boolean just for this?

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export interface IGlobalNavigationLogo extends IBaseGlobalNavigationItem {}
+
+interface MenuItem extends Omit<MenuItemType, 'key'> {
+  children?: MenuItem[]
+}
+
+export interface IGlobalNavigationTool extends IBaseGlobalNavigationItem {
+  type: 'menu' | 'link'
```
tibuurcio, line 30: I think we should type this more like `parent | children` cause our menu's can be links as well and vice versa

src/components/navigation/GlobalNavigation/WorkspaceSelector.tsx
```diff
@@ -0,0 +1,118 @@
...
+            onClick: workspace.onClick,
+          })
+        })
+      })
+
+      return total
+    }, [])
+  }
+
+  function onSearch(e: React.ChangeEvent<HTMLInputElement>): void {
```
tibuurcio, line 101: We might want to consider [debouncing](https://lodash.com/docs/4.17.15#debounce) the onSearch function just a small amount for performance reasons since this updates on keystroke.

tibuurcio:
@nastyastavitskaya should be interesting for you to take a look at how the nav component works so I've marked you as reviewer here

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
jared-dickman: hm i dont think "our menus can be links" is correct, in the current design, you can not click on a menu item, and a link wont have a menu expansion. 

infact the way its built now, the children of a menu can be either `type link` or `type menu` so theres a bit of nesting going on

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
jared-dickman: consider renaming Create to something more generic

design/LightTheme.ts
jared-dickman: considerAdding rounding function in getGlobalToken

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
jared-dickman: probably

jared-dickman:
note - i brought in the latest dependabot branch, so we can qa the new packages for a few days while product testing

.gitignore
```diff
@@ -89,7 +89,6 @@ out
 
 # Nuxt.js build / generate output
 .nuxt
-dist
```
jared-dickman, line 92: todo: this is for development only

src/components/data-display/Avatar/Avatar.stories.tsx
```diff
@@ -18,7 +15,7 @@ const meta: Meta<typeof Avatar> = {
   args: {
     alt: 'User Avatar',
     gap: 4,
-    icon: <Icon icon={faUser} />,
```
jared-dickman, line 21: with the removal of font awesome, and the pending of a new Icon component, all story icons have temporarily been replaced with a generic icon [mp logo]

src/utils/utils.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { getInitials } from './utils'
```
jared-dickman, line 1: [TD](https://mparticle-eng.atlassian.net/browse/UNI-191) item to get this working

src/utils/utils.css
```diff
@@ -0,0 +1,3 @@
+.u-display-none {
```
jared-dickman, line 1: consider how utils might be exported downstream from the aquarium

tibuurcio (approved):
Let's go!

PR #99, opened 2024-02-08
feat: icons

jared-dickman:
## Instructions

## Summary
- Install font-awesome pro icons!

## Testing Plan

- Closes https://go.mparticle.com/work/UNI-136


package.json
```diff
@@ -20,9 +20,7 @@
     "dist/_variables.css"
   ],
   "dependencies": {
-    "@fortawesome/fontawesome-svg-core": "6.5.1",
-    "@fortawesome/free-solid-svg-icons": "6.5.1",
-    "@fortawesome/react-fontawesome": "0.2.0"
+    "@awesome.me/kit-f3b115a3e2": "^1.0.7"
```
jared-dickman, line 23: i think a `^` here is ok, because this kit-id is linked to our org account and controlled by the design team. when they push updates to our kit icons, we should be able to get them without manually updating the package.json

package.json
```diff
@@ -20,9 +20,9 @@
     "dist/_variables.css"
   ],
   "dependencies": {
-    "@fortawesome/fontawesome-svg-core": "6.5.1",
-    "@fortawesome/free-solid-svg-icons": "6.5.1",
-    "@fortawesome/react-fontawesome": "0.2.0"
+    "@awesome.me/kit-f3b115a3e2": "^1.0.7",
```
jared-dickman, line 23: it probably makes sense to keep the `^` on this package, since this is our personal kit-code and maintained by the design team. this means when they push new icon updates we dont have to change the package.json

.gitignore
```diff
@@ -50,6 +50,7 @@ web_modules/
 
 # Optional npm cache directory
 .npm
+.npmrc
```
tibuurcio, line 53: Should this be here? Since we are commiting the .npmrc file

package.json
tibuurcio: Interesting!

tibuurcio (approved):
Looking good!

.gitignore
jared-dickman: i think yes, the idea is we want to commit the npmrc with this pr, and then not commit changes subsequently

nastyastavitskaya (approved):

PR #105, opened 2024-02-13
chore: global nav prep / cleanup

jared-dickman:
## Instructions

## Summary
- Bringing in changes from the global nav which can be merged prior to [the pr](https://github.com/mParticle/aquarium/pull/36)
- this will reduce the review and testing effort for global nav

## Testing Plan


src/styles/readme.md
tibuurcio, line 1: Nice! ✨ 📝

src/utils/debounce.ts
```diff
@@ -0,0 +1,11 @@
...
+  let timeoutId: number
+
+  return (...args: any[]) => {
+    window.clearTimeout(timeoutId)
+    timeoutId = window.setTimeout(() => {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+      callback(...args)
+    }, wait)
+  }
+}
```
tibuurcio, line 11: We had a sync discussion about including this instead of something like @lodash/debounce. This change makes it a little bit better on the eyes.

```suggestion
export const debounce = (callback: (...args: unknown[]) => void, wait: number): ((...args: unknown[]) => void) => {
  let timeoutId: number

  return (...args: unknown[]) => {
    window.clearTimeout(timeoutId)
    timeoutId = window.setTimeout(() => {
      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
      callback(...args)
    }, wait)
  }
}
```

tibuurcio (approved):
Thanks for splitting the changes in this PR from the other one 👍🏼

src/utils/debounce.ts
jared-dickman: having unkown as the first type caused type errors at implementation. 
i've [made some changes](https://github.com/mParticle/aquarium/blob/1dcdf0e229ea1e3a2bc8ef3d6cae09d290bc50cb/src/utils/utils.ts#L15-L14) and moved it into a utils file in the followup pr

PR #106, opened 2024-02-13
build: Applies vite default configs for bundling as a module and commonjs

tibuurcio:
## Summary
- Applies [vite default configs](https://vitejs.dev/guide/build.html#library-mode:~:text=Running%20vite%20build%20with%20this%20config) to build towards shipping libraries and also includes `type: "module"` in package.json for the same purpose.

I will push this changes to #36 as well but making them here to document the process and related links.

## Testing Plan

Tested a bunch of [different output formats](https://rollupjs.org/configuration-options/#output-format) but those ones were the ones that worked better without having to ship `antd` inside the bundle.

- Closes https://go.mparticle.com/work/UNI-170


jared-dickman:
seems to be working in ind

jared-dickman (approved):

PR #123, opened 2024-03-01
fix: Input now support refs

tibuurcio:
# Summary
This PR adds support for using Refs in our Input component the same way Antd's native Input supports it. With this we are able to support implementations that need to control input focus.

[Implementation was based on the one from Antd](https://github.com/ant-design/ant-design/blob/7abae5e973e50143b2b1b52731faea0eb6b58121/components/input/index.ts)

## Testing Plan
- [X] Was this tested locally? If not, explain why.
- Run storybook and run tests

https://github.com/mParticle/aquarium/assets/5959178/7523b676-efcb-48d7-abec-6f86b632dfa3

## Reference Issue (For mParticle employees only.  Ignore if you are an outside contributor)
- Closes https://go.mparticle.com/work/UNI-225

PR #124, opened 2024-03-04
feat: global navigation and styling

jared-dickman:
## Instructions

## Summary

## Testing Plan


.github/workflows/release.yml
```diff
@@ -4,15 +4,9 @@ on:
   workflow_dispatch:
 
 jobs:
-  # release is done from main branch.
```
jared-dickman, line 7: **Issue** lets not remove this 

infact probably we should have some protections [perhaps code owners] to ensure this is not removed

.gitignore
```diff
@@ -89,7 +89,6 @@ out
 
 # Nuxt.js build / generate output
 .nuxt
-dist
```
jared-dickman, line 92: these changes also should probably never go into main

we still dont have a good/scalable way for the rest of the team to use branches still in development, in the platforms

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mparticle/aquarium",
-  "version": "1.1.1",
+  "version": "1.6.2",
```
jared-dickman, line 3: this should _ideally_ be an automated part of the release

package.json
```diff
@@ -43,6 +43,7 @@
     "@typescript-eslint/eslint-plugin": "6.19.0",
     "@vitejs/plugin-react": "4.2.1",
     "concurrently": "8.2.2",
+    "cz-conventional-changelog": "^3.3.0",
```
jared-dickman, line 46: do we want the `^` here??

release.config.cjs
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  branches: ['main'],
+  branches: ['main', 'next'],
```
jared-dickman, line 2: **issue** in production, only main should be releasable

src/components/data-entry/Mentions/Mentions.stories.tsx
```diff
@@ -67,10 +67,10 @@ export const Primary: Story = {
       options: ['outlined', 'borderless', 'filled'],
     },
   },
-// @ts-ignore
+  // @ts-expect-error TODO
```
jared-dickman, line 70: can you add more clarity around what is expected to be done?

src/components/icons/index.ts
```diff
@@ -0,0 +1,41 @@
+import ChartLineIcon from 'src/assets/svg/chart-line.svg?react'
```
jared-dickman, line 1: **suggestion** add a readme in this directory so other engineers know the complete process to get icons from design into the platform 
[that it must be exported here _and_ in src/components/index]

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
```diff
@@ -415,6 +415,38 @@ export const MP: Story = {
   },
 }
 
+thousandOrgs.forEach(org => {
```
jared-dickman, line 418: **thought** some of the data in the thousandOrgs seems not the best/cleanest to add to the storybook..

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
```diff
@@ -1,25 +1,26 @@
...
+import { useRef, useState } from 'react'
 import { useCallback } from 'react'
 import { useEffect } from 'react'
 import { useMemo } from 'react'
-import { useRef } from 'react'
 import { debounce } from 'src/utils/utils'
 import { getInitials } from 'src/utils/utils'
 
+// Need to make our Input component comply with forwardRef to be able to import it from src/components
+// Couldn't make it work as of right now
```
jared-dickman, line 19: **suggestion** add a `todo`?

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
```diff
@@ -123,37 +62,75 @@ export function WorkspaceSelector(props: IWorkspaceSelectorProps) {
...
-  const items: IMenuProps['items'] = [
-    {
-      key: 'WorkspaceSelector',
-      icon: <Avatar className="workspaceSelector__avatar">{getInitials(activeWorkspace?.label)}</Avatar>,
-      popupClassName: 'workspaceSelector',
-      children: menuChildren,
-    },
-  ]
-
+    .find(workspaceCandidate => workspaceCandidate.isActive)!
```
jared-dickman, line 65: **question** what does a `!` at the end do?

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
```diff
@@ -123,37 +62,75 @@ export function WorkspaceSelector(props: IWorkspaceSelectorProps) {
...
-      // openKeys={['WorkspaceSelector']} // testing only
-      className="globalNavigation__menu globalNavigation__item globalNavigation__item--workspaceSelector"
-      items={items}
-      onOpenChange={clearSearch}
-      expandIcon={null}
-    />
+    <Popover
+      placement="right"
+      // Can't seem to find a way to access it via className
+      overlayInnerStyle={{ padding: 4 }}
```
jared-dickman, line 83: this would be a usecase for using variables.ts, once stylel-dictionary is exporting both css and ts files

src/components/navigation/GlobalNavigation/WorkspaceSelector/workspace-selector.css
```diff
@@ -1,91 +1,110 @@
...
+.workspaceSelector__noResults {
+  height: 330px;
+  pointer-events: none;
+}
 
-.workspaceSelector__search {
-    position: sticky !important;
-    bottom: 40px;
-    z-index: 1;
-
```
jared-dickman, line 10: sad that it removed all whitespace. its valuable to separate ideas

src/components/navigation/GlobalNavigation/WorkspaceSelector/workspace-selector.css
```diff
@@ -1,91 +1,110 @@
...
-    width: 100% !important;
-    padding: 0 var(--padding-xxs) !important;
-    margin: 0 !important;
+.workspaceSelector__noResults .ant-result-title {
+  font-size: var(--font-size);
+}
 
-    background-color: white !important;
+.workspaceSelector__menuItem {
+  display: flex;
```
jared-dickman, line 20: this looks like the `Center` component, rewritten in css 🤔 
wonder if the center component is slower, since its in ant/ts...

src/utils/utils.spec.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-undef */
```
jared-dickman, line 1: i would prefer using disable-next-line, or adding eslint-enable at the bottom of the file

package.json
tibuurcio: For sure! I've seen [this package](https://github.com/semantic-release/git) that could be used for that but haven't dived deeper. Create [a ticket for this.](https://mparticle-eng.atlassian.net/browse/UNI-227)

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
tibuurcio: Makes sense. Also #123 fixes this so might as well update it on this PR

src/components/navigation/GlobalNavigation/WorkspaceSelector/workspace-selector.css
tibuurcio: Agree.. I tried using it and somehow it gave different results. Have to take another look into it.

.github/workflows/release.yml
tibuurcio: I think it would be benefitial if we had a flow where releasing from next is allowed. This could be useful for users to test our next releases without us having to publish stuff to the main. Also it seems to be supported by both npm and semantic-release as a "native" thing. See [this docs](https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration#pushing-to-a-release-branch)

image

Thoughts?

We would have to figure it out how it would work though

tibuurcio: Also, for testing purposes I think we could have beta/alpha versions. Things like `v1.7.3-beta.0`. This helps not releasing so many versions.

.gitignore
tibuurcio: For sure

package.json
tibuurcio: Probably not, let's change that.

release.config.cjs
tibuurcio: [See this comment](https://github.com/mParticle/aquarium/pull/124#discussion_r1512845374)

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
tibuurcio: Yeah... I was just trying to replicate the data from production but we should probably add some generator function to generate data for this.

package.json
```diff
@@ -90,5 +91,10 @@
     "*.{json,yml,md}": [
       "prettier --write"
     ]
+  },
+  "config": {
+    "commitizen": {
+      "path": "./node_modules/cz-conventional-changelog"
+    }
```
tibuurcio, line 98: We need documentation for that.

src/components/data-entry/Mentions/Mentions.stories.tsx
tibuurcio: I'm not sure what's going on here :/ eslint complains about usage of @ts-ignore and suggests @ts-expect-error instead. Taking a look at this.

tibuurcio: Found and fixed. `status` and `placement` should be inside `argTypes`

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
```diff
@@ -415,6 +415,38 @@ export const MP: Story = {
...
+          management={mpManagement}
+          orgs={thousandOrgs}
+          onMpHomeClick={() => {
+            alert('going to overview map')
+          }}
+        />
+      </div>
+    )
+  },
+}
```
tibuurcio, line 453: I thought of extracting this to another story to isolate MPOrgSwitcher behaviour but also this is currently tied to the GlobalNavigation component so it makes sense to live here.

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
tibuurcio: Active workspace can be null if we don't find an active one and thus also the need for `// eslint-disable-next-line @typescript-eslint/no-non-null-assertion` in order to be able to use the `!` here. This part likely need some redesign cause there's also a comment for not needing to be calculated on every render.

Think we can address this at [UNI-185](https://mparticle-eng.atlassian.net/browse/UNI-185)

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
tibuurcio: Added a comment for that

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
```diff
@@ -123,37 +62,75 @@ export function WorkspaceSelector(props: IWorkspaceSelectorProps) {
...
+          </div>
+
+          {hasNoResults ? (
+            <WorkspaceSearchLabel />
+          ) : (
+            <ul className="workspaceSelector__itemsList">
+              {menuItems.map(item => (
+                <li key={item.key} className={item.className} onClick={item.onClick}>
+                  {item.label}
+                </li>
```
tibuurcio, line 111: We should improve the onClick handler here by providing the event tha triggered it. Also, types are currently a missmatch

image

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
```diff
@@ -123,37 +62,75 @@ export function WorkspaceSelector(props: IWorkspaceSelectorProps) {
...
+              value={searchTerm}
+              ref={inputRef}
+              onClick={e => {
+                e.preventDefault()
+                e.stopPropagation()
+              }}
+            />
+          </div>
+
+          {hasNoResults ? (
```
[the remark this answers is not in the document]
tibuurcio: Agree. I'm extracing the whole content to a new component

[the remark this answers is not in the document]
tibuurcio: Extracting it make parens not needed
```jsx
{hasNoResults ? <WorkspaceSearchLabel /> : <WorkspaceSelectorContentItems menuItems={menuItems} />}
```

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelectorItems.d.ts
```diff
@@ -1,24 +1,23 @@
...
-  workspaces: INavigationWorkspace[];
-  onClick?: (info: IMenuInfo) => void;
+  workspaces: INavigationWorkspace[]
+  onClick?: () => void
 }
 
 export interface INavigationWorkspace extends IWorkspaceSelectorItem {
   isActive: boolean
-  onClick?: (info: IMenuInfo) => void
+  onClick?: () => void
```
tibuurcio, line 20: We need a better interface for this handler.

src/components/navigation/GlobalNavigation/WorkspaceSelector/workspace-selector.css
tibuurcio: Could we use comments on those cases? Feel if there's need to separate ideas in the same file, comments are a good use for that.

src/utils/utils.spec.ts
tibuurcio: solved by adding `jest: true` to `envs`  in `.eslintrc.cjs`

.gitignore
```diff
@@ -138,4 +137,6 @@ out/
 test/reports
 test/test-bundle.js.map
 test/stub/test-stub-bundle.js
-**/dist
\ No newline at end of file
+
+#dist/
+#**/dist
```
tibuurcio, line 142: With `next` being releasable we can publish it and install from `@mparticle/aquarium@next`. Ideally we should be able to test locally but having it from `next` seems better than our current approach of pushing dist/

package.json
tibuurcio: Probably should also not allow commits without the format cause semantic-release ignores commits without it and it messes up with the release process

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
tibuurcio, line 1: @gabrielmarquesf Might be a good place for your performance suggestions!

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
```diff
@@ -175,17 +131,27 @@ export function WorkspaceSelector(props: IWorkspaceSelectorProps) {
...
+      id: item.id,
+      key: item.key,
+      onClick: item.onClick,
+    }))
   }
 
-  function onSearch(e: React.ChangeEvent<HTMLInputElement>): void {
+  function onSearch(e: ChangeEvent<HTMLInputElement>): void {
     const newSearchTerm = e.target.value.toLowerCase()
     setSearchTerm(newSearchTerm)
```
gabrielmarquesf, line 156: When we set the state here we are re-rendering the component, and when we set the `currentFilteredOrgs` state after 200ms we re-render again.
My suggestion is to remove this `searchTerm` state, and let this input to be uncontrolled. Because I can't see usage for this state here.
An alternative would be to use `memo()` in the `WorkspaceSelectorContentItems` to avoid re-render a long list if the prop is the same.

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
gabrielmarquesf: I see it's quite fast now already :)

gabrielmarquesf (commented):
Left some optional suggestions for the `WorkspaceSelector.ts` comp

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
```diff
@@ -175,17 +131,27 @@ export function WorkspaceSelector(props: IWorkspaceSelectorProps) {
...
     }, [])
+
+    // prevent attributes to end up in the HTML
+    return items.map(item => ({
+      type: item.type,
+      className: item.className,
+      label: item.label,
+      id: item.id,
+      key: item.key,
+      onClick: item.onClick,
```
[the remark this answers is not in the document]
tibuurcio: Good catch! We might not need it anymore.

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
tibuurcio: Tried removing `searchTerm` but we need it to determine if we are searching or not.
```
const hasNoResults = !!searchTerm && !currentFilteredOrgs.length
```

We are currently memo'ing the menuItems so I think we are covered 🙏🏼

tibuurcio (approved):

PR #139, opened 2024-03-14
Fix: add check next to active workspace

jared-dickman:
## Instructions

## Summary
- Add checkmark next to active workspace, per the mocks
- Add wrapping in un-breakable long workspace/account/org names, to avoid horizontal scroll bar

## Testing Plan
- testing in storybook for all GlobalNavigation stories

- Closes https://go.mparticle.com/work/UNI-223

## Screenshots
image

image

image


tibuurcio (approved):
Nice work!

PR #148, opened 2024-03-15
fix: active workspace style

jared-dickman:
## Instructions

## Summary
- fix check missing in workspace selector

## Testing Plan


tibuurcio (approved):
Tested locally no mP. Looks good!

PR #149, opened 2024-03-15
refactor: running prettier in the codebase with a new config

tibuurcio:
## Summary
- Changing prettier `bracketSameLine` to `true` to make more sense with our `printWidth` config

image

## Testing Plan
- [X] Was this tested locally? If not, explain why.

PR #151, opened 2024-03-16
fix: Uses anchor links for menu items in the nav component

tibuurcio:
## Summary
This PR changes the way we interact with menu items by allowing the `href` property for menu items and wrapping those in anchor elements. It does that by:
- Making `href` required for `type: 'link'` elements
- Making our item types simpler since we don't support subchilds anymore
- Removing `cursor: pointer` from Menu elements

## Testing Plan
- Tests are implemented [within Storybook](https://storybook.js.org/docs/writing-stories/play-function) although I couldn't find a way to interact with the Menu Popover element. There's [this issue](https://github.com/storybookjs/storybook/issues/16971) on that but I couldn't make it work.

- Closes https://go.mparticle.com/work/UNI-228

This should also go after #149 and #150 cause there's some commits from those PR's here.


src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
```diff
@@ -40,46 +40,64 @@ const defaultLogo: IGlobalNavigationLogo = {
...
   {
     label: 'Tool 1',
     isActive: true,
     icon: <ShieldKeyholeIcon />,
     type: 'menu',
-    children: [{ label: 'option 1' }, { label: 'option 2' }, { label: 'option 3' }],
+    children: [
+      { label: 'option 1', href: '/' },
+      { label: 'option 2', href: '/' },
+      { label: 'option 3', href: '/' },
```
tibuurcio, line 53: `href`'s are now required for menu links. Since we are currently not supporting subchilds this could be done on a single level basis.

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
```diff
@@ -40,46 +40,64 @@ const defaultLogo: IGlobalNavigationLogo = {
...
+    ],
   },
   {
     label: 'Tool 3',
     icon: <ConnectionsIcon />,
-    type: 'link',
+    href: '/',
```
tibuurcio, line 69: Default will be `type: 'link'` so no need to declare it anymore.

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
```diff
@@ -25,13 +25,13 @@ import { CloudIcon } from 'src/components'
 import { Badge } from 'src/components/data-display/Badge/Badge'
 import { type INavigationOrg } from 'src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelectorItems'
 import { type IGlobalNavigationLogo } from 'src/components/navigation/GlobalNavigation/GlobalNavigationItems'
-import { type IGlobalNavigationManagement } from 'src/components/navigation/GlobalNavigation/GlobalNavigationItems'
-import { type IGlobalNavigationTool } from 'src/components/navigation/GlobalNavigation/GlobalNavigationItems'
+import { type IGlobalNavigationItem } from 'src/components/navigation/GlobalNavigation/GlobalNavigationItems'
```
tibuurcio, line 28: Merged items in a single type

src/components/navigation/GlobalNavigation/NavigationList.tsx
```diff
@@ -1,63 +1,67 @@
...
+  ]
 
-    children.unshift({ label: item.label, type: 'group', key: item.label + '_groupTitle' })
+  if (item.type === 'menu') {
+    children.push(
+      ...item.children.map((linkItem, j) => ({
+        ...linkItem,
+        expandIcon: linkItem.isNestedMenu ? true : null,
+        key: `${String(linkItem.label)}${j}`,
+        label: <a href={linkItem.href}>{linkItem.label}</a>,
```
tibuurcio, line 46: Here's the fix

tibuurcio, line 46: Here's the fix 🙏🏼

src/components/navigation/GlobalNavigation/global-navigation.css
```diff
@@ -96,6 +100,10 @@
   width: 100%;
 }
 
+.navigationCreate__popupButton {
+  cursor: revert !important;
+}
```
tibuurcio, line 105: Gotta check if that's correct for indicative @jared-dickman

.storybook/main.ts
```diff
@@ -6,7 +6,10 @@ import { withoutVitePlugins } from '@storybook/builder-vite'
 type StorybookVitePlugins = { plugins: (PluginOption[] | Plugin)[] }
 
 const config: StorybookConfig & StorybookVitePlugins = {
-  framework: '@storybook/react-vite',
+  framework: {
+    name: '@storybook/react-vite',
+    options: {},
```
jared-dickman, line 11: is this needed?

package.json
```diff
@@ -29,18 +29,16 @@
...
-    "@storybook/addon-onboarding": "1.0.11",
-    "@storybook/blocks": "7.6.14",
-    "@storybook/builder-vite": "7.6.14",
-    "@storybook/jest": "0.2.3",
-    "@storybook/react": "7.6.10",
-    "@storybook/react-vite": "7.6.10",
-    "@storybook/test": "7.6.10",
-    "@storybook/test-runner": "0.16.0",
-    "@storybook/testing-library": "0.2.2",
+    "storybook": "8.0.0",
```
jared-dickman, line 32: :woohoo:

src/components/navigation/GlobalNavigation/NavigationList.tsx
jared-dickman: so it still supports onClick via `...linkItem`?
do you think we should have a check before rendering the `a`, or before adding the `href` attribute, that its defined and a good/valid link?

not every item in analytics acts as an a with a valid href 😩

jared-dickman: maybe thats new "correct" in the new world, but would have to chat with product. for example Explore Demo would is one theyve always wanted to open in a new tab, so it doesnt clear your current session context

src/components/navigation/GlobalNavigation/global-navigation.css
jared-dickman: probably best to use the value we expect instead of `revert`? thinking making this `cursor` would be "correct"

src/components/navigation/GlobalNavigation/NavigationList.tsx
tibuurcio: Still supports onClick, yes! Good point about not all items supporting href's from the beginning. I think a check it's important here, but we will also have to make it optional so the types reflect it better. I'm gonna use the ind story to setup an example of this. Thanks for the callout

src/components/index.ts
```diff
@@ -64,7 +64,6 @@ export { Breadcrumb, type IBreadcrumbProps } from './navigation/Breadcrumb/Bread
 export { Pagination, type IPaginationProps } from './navigation/Pagination/Pagination'
 export {
   Menu,
-  MenuItem,
```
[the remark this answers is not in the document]
tibuurcio: It makes the migration to Storybook 8 fails, not sure why 🤔 
Also we don't use it currently.

.storybook/main.ts
tibuurcio: Storybook migration added it. Not sure, probably not.

src/components/navigation/GlobalNavigation/NavigationList.tsx
tibuurcio: Also `href` is ending up in the `li` elements right now :/

image

tibuurcio: Added a check for the `href` prop and made it optional

.storybook/main.ts
tibuurcio: Required
image

src/components/navigation/GlobalNavigation/NavigationList.tsx
nastyastavitskaya: same is true for cortex. it used to use either `<a href=""` for external navigation or  `<NavLink to="">` for internal. Will it remain the same as it is now with onClick for internal navigation?

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
jared-dickman: i feel like we wont have an href for every single menu option in analytics.
for example, how do we expect it to work for items that open in a new tab?

jared-dickman (approved):

PR #162, opened 2024-03-20
feat: Allows publishing on dev branch/distribution channel

tibuurcio:
## Summary
This PR introduces a few things in order to start our path into a better release process for testing.
- Creates the concept of the `@dev` release branch, which will follow semver beta versioning, like `@1.8.3-dev.0`. This will allows us to have proper versions published to npm to test in QA environments. After dev is merged into main, semantic-release will perform the release of `@1.8.3` on the `@latest` distribution channel.
  - If we ever release `@1.8.3` before finishing up the work on the dev branch, we just have to merge main -> dev and the dev release will become `@1.8.4-dev.0`. Changelog should also help us with keeping track of what's in each release.
- Removes the tokens-to-css github action and runs it on the `pre-commit` hook with the help of husky. This will prevent commits whenever we make a push to the repo.
- Uses the @semantic-release/git plugin to update the package version automatically.
- Uses the @semantic-release/changelog plugin to update CHANGELOG automatically.
- Uses commitlint and husky's `commit-msg` hook to make sure commits messages adhere to the expected format
- Docs changes

## Testing Plan
- [Here's the release action](https://github.com/mParticle/aquarium/actions/runs/8361941915/job/22891272808)
- [Here's the package.json update commit](https://github.com/mParticle/aquarium/pull/162/commits/b20e21bb9544275894302fac1b62f892191f5c3c)
- [Here's the CHANGELOG file](https://github.com/mParticle/aquarium/blob/dev/CHANGELOG.md)

- Closes https://go.mparticle.com/work/UNI-227
- Closes https://go.mparticle.com/work/UNI-294
- Closes https://go.mparticle.com/work/UNI-309
- Closes https://go.mparticle.com/work/UNI-308


.husky/pre-commit
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env sh
 . "$(dirname "$0")/_/husky.sh"
 
+npm run tokens-to-css
```
jared-dickman, line 4: **praise** looks much better!

jared-dickman:
great work here, as usual :smile:

.github/workflows/reusable-workflows.yml
```diff
@@ -4,9 +4,6 @@ on:
   pull_request:
 
 jobs:
-  pr-branch-check-name:
```
jared-dickman, line 7: hmmm 🤔

package.json
```diff
@@ -29,7 +29,10 @@
     "react-dom": ">=17.0.2"
   },
   "devDependencies": {
-    "storybook": "8.0.0",
+    "@commitlint/cli": "^19.2.1",
+    "@commitlint/config-conventional": "^19.1.0",
```
jared-dickman, line 33: whats the thought behind including the `^` in this package specifically ?

tibuurcio: It's the opposite of it. Not thinking about it 😆 
Changing this, good catch.

.github/workflows/reusable-workflows.yml
tibuurcio: it fails for the dev branch :/ 
we can have a PR branch check in the future it just need to be something customized

jared-dickman: yes thats what i thought

is it a bad idea to use, instead of `dev`, `build/dev`?

otherwise we can copy the functionality from `mParticle/mparticle-workflows/.github/workflows/pr-branch-check-name.yml@stable` and modify it for use in this repo

tibuurcio: I think the downside is diverging from conventions, cause of the reasoning behind `build/*` branches/commits from conventional commits: `build: Changes that affect the build system or other dependencies`.

tibuurcio: I would be okay with copying the functionality for now

tibuurcio: Added the new workflow with the dev branch addition

.github/workflows/pr-branch-check-name.yml
```diff
@@ -0,0 +1,58 @@
...
+    name: 'Check PR for semantic branch name'
+    if: ${{ github.actor != 'dependabot[bot]' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Set PR branch validity'
+        id: is-semantic
+        if: >
+          startsWith(github.event.pull_request.head.ref, 'feat/') ||
+          startsWith(github.event.pull_request.head.ref, 'fix/') ||
+          startsWith(github.event.pull_request.head.ref, 'perf/') ||
```
jared-dickman, line 17: these dont line up 100% with the [release.config](https://github.com/mParticle/aquarium/blob/7863d9acc33d9cc3447c598672a31d5e6bbedf5a/release.config.cjs#L20) file, but probably should

for example, i dont recognize `perf`

.github/workflows/pr-branch-check-name.yml
```diff
@@ -0,0 +1,58 @@
+﻿name: 'Branch Name PR Check'
```
jared-dickman, line 1: should we make this a required check once its merger in?

.github/workflows/pr-branch-check-name.yml
tibuurcio: it should, yes

.github/workflows/pr-branch-check-name.yml
jared-dickman: the repo settings [branch protections] yes

jared-dickman (approved):

PR #163, opened 2024-03-20
feat: focus workspace on icon click

jared-dickman:
## Instructions

## Summary
- currently the workspace selector focuses the input on open of the popover, but users may feel it is natural still to click the icon. this unfocuses the input
- this pr allows clicking the icon to focus, making it feel less broken

## Testing Plan
- test in storybook 


tibuurcio (approved):
Nice catch

PR #166, opened 2024-03-21
fix: nav link target

jared-dickman:
## Instructions

## Summary
- Allow nav links to open in a new tab

## Testing Plan


src/components/navigation/GlobalNavigation/NavigationItem.tsx
```diff
@@ -12,6 +12,7 @@ export interface INavigationItemProps {
   isActive?: boolean
   onClick?: (e: MouseEvent) => void // link only
   href?: string // link only
+  hrefTarget?: '_self' | '_blank' // link only
```
tibuurcio, line 15: Could we make this into the href property itself? Something [like this](https://www.gabe.pizza/notes-on-component-libraries/#:~:text=It%20does%20not%20make%20sense%20for%20the%20interface%20to%20have%20two%20props%2C%20label%20and%20labelTooltip%2C%20because%20the%20tooltip%20won%E2%80%99t%20show%20without%20the%20label.%20They%20should%20be%20merged%20into%20a%20single%20prop%20to%20capture%20this%20requirement%3A) example. We won't need hrefTarget to be defined without href so it seems like a good case for that.

jared-dickman: interesting! i like it, but i dont like as much the `normalizeLabel` method they need to create (thats not mentioned anywhere else)

what about using `hrefOptions?: { link: string, target?: '_self' | '_blank' }`

tibuurcio: Seems okay, I would only change `link` for `href` maybe, to keep it consistent with the HTML option. Not ideal having `hrefOptions` and `href` again inside of it but would be ok with that.

tibuurcio: That's also a breaking change for the current platforms, so one drawback of using it would be that. I think that's fine for the moment we are but future updates should account for that.

jared-dickman: this is true its a breaking change for the prop, but, none of the platforms currently use this [analytics is on 1.8.2, cdp is on 1.7.9, cortex on 1.7.9] so not a real issue for us now

tibuurcio (approved):

PR #167, opened 2024-03-22
fix: sub menu links should open the correct herf

jared-dickman:
## Instructions

## Summary
- this fixes a ts error in overlooked in the previous pr
- allowing links inside of menus to open correctly

## Testing Plan


src/utils/utils.tsx
```diff
@@ -48,3 +50,16 @@ export function hasImageAtSrc(src: string, hasImageSetter?: (hasImage: boolean)
     img.src = src
   })
 }
+
+export type hrefOptions = { href: string; hrefTarget?: '_self' | '_blank' }
```
tibuurcio, line 54:
```suggestion
export type HrefOptions = { href: string; hrefTarget?: '_self' | '_blank' }
```

tibuurcio: For consistency

jared-dickman: or should it be `IHrefOptions`??

jared-dickman: guess not since its not an interface?

tibuurcio: Yeah I mean.. I was never a fan of the `I` prefix even for interfaces. Too much OO for me 😅

tibuurcio (approved):

src/utils/utils.tsx
jared-dickman: what does OO stand for??

jared-dickman: let the record show, i am always a fan of less typing, as long as it doesnt obscure the meaning

PR #168, opened 2024-03-22
style: Update Global Nav scroll styling

nastyastavitskaya:
## Summary

- We have an issue with the scroll bar in mParticle, where it cuts the "Data Platform" text. Also, mP codebase has some global scroll styling, which leads to uneven experience with the component in 3 apps. This PR contains some custom styling to the scroll bar to fix these issues.

## Testing Plan

- Run global navigation stories and check the styling of the scroll bar.

- Closes https://mparticle-eng.atlassian.net/browse/UNI-222

## Screenshots
Browsers that support `scrollbar-*` properties
Screen Recording 2024-03-21 at 3 19 49 PM

Browsers that use `::-webkit-scrollbar-*` pseudo-elements
Screen Recording 2024-03-21 at 3 22 21 PM


src/components/navigation/GlobalNavigation/global-navigation.css
```diff
@@ -3,6 +3,9 @@
   --nav-item-width: 82px;
   --nav-width: 90px;
   --suite-logo-height: 82px;
+  --color-scrollbar-handle: rgb(0 0 0 / 15%);
+  --color-scrollbar-track: transparent;
+  --size-scrollbar-handle: 6px;
```
nastyastavitskaya, line 8: **note:** colors are taken from Indicative codebase, however the size I had to tweak. Indicative use `8px`, but with this value the "Data Platform" text still gets cut. 4 looks too thin, so 6 looks like the best alternative

src/components/navigation/GlobalNavigation/global-navigation.css
```diff
@@ -22,6 +25,32 @@
   overflow: hidden auto;
 }
 
+/* For browsers that support `scrollbar-*` properties */
+@supports (scrollbar-color: auto) {
```
nastyastavitskaya, line 29: **learning:** using `scrollbar-color` and `scrollbar-width` is the more modern way to handle scroll styling, but not all browsers support it. If supported they will override `::-webkit-scrollbar-*` styling. Check the screenshot section in PR to see the differences.  `::-webkit-scrollbar-*` is what we currently have in mP

jared-dickman: cool!

src/components/navigation/GlobalNavigation/global-navigation.css
```diff
@@ -22,6 +25,32 @@
...
+@supports selector(::-webkit-scrollbar) {
+  .globalNavigation::-webkit-scrollbar {
+    /* stylelint-disable-next-line property-no-vendor-prefix */
+    -webkit-appearance: none;
+    width: var(--size-scrollbar-handle);
+    height: var(--size-scrollbar-handle);
+  }
+
+  .globalNavigation::-webkit-scrollbar-track {
+    background: var(--color-scrollbar-track);
```
jared-dickman, line 46: **suggestion** can we use background-color here instead??

`background` and `background-color` will result in the same styles in most cases, where other background rules are not present, but its a little safer and more explicit to use `background-color` in the case that any other `background` rules are applied. 


https://stackoverflow.com/questions/10205464/what-is-the-difference-between-background-and-background-color
```
background will supercede all previous background-color, background-image, etc. specifications. It's basically a shorthand, but a reset as well.
```

nastyastavitskaya: great point, let's do it 👍

jared-dickman (approved):
code looks good, tested locally in storybook

PR #169, opened 2024-03-22
fix: Input now support refs

tibuurcio:
# Summary
This PR adds support for using Refs in our Input component the same way Antd's native Input supports it. With this we are able to support implementations that need to control input focus.

[Implementation was based on the one from Antd](https://github.com/ant-design/ant-design/blob/7abae5e973e50143b2b1b52731faea0eb6b58121/components/input/index.ts)

## Testing Plan
- Run storybook and run tests

https://github.com/mParticle/aquarium/assets/5959178/7523b676-efcb-48d7-abec-6f86b632dfa3

- Closes https://go.mparticle.com/work/UNI-225


src/components/data-entry/Input/Input.tsx
```diff
@@ -2,20 +2,31 @@ import { Input as AntInput } from 'antd'
...
 
+const InternalInputWithRef = forwardRef(InternalInput)
+const Input = InternalInputWithRef as CompoundedComponent
+
 Input.Group = AntInput.Group
 Input.Password = AntInput.Password
 Input.Search = AntInput.Search
 Input.TextArea = AntInput.TextArea
+
+export { Input, type InputRef, type IInputProps }
```
jared-dickman, line 32: can we move this to the top of the file for consistency?

tibuurcio: yes, also looking into an eslint rule for this

tibuurcio: There's no rule for that :( 
image

jared-dickman (approved):
nice one

PR #182, opened 2024-03-28
feat: icon component

jared-dickman:
## Instructions

## Summary
- Create reusable component for Icons
- All props should be typed, limiting the size/color/name of the icon 
- This matches what the design team expects

## Testing Plan

- Closes https://go.mparticle.com/work/UNI-226


src/components/general/Icon/Icon.tsx
```diff
@@ -1,11 +1,24 @@
...
+import { Icons } from 'src/constants/Icons'
+import './icon.css'
+
+type IconColor = 'primary' | 'success' | 'warning' | 'error' | 'info' | 'white' | 'text'
+type IconSize = 'xxl' | 'xl' | 'lg' | 'md' | 'ms' | 'sm' | 'xs' | 'xxs'
+
+export interface IIconProps {
+  name: keyof typeof Icons
+  color?: IconColor
+  size?: IconSize
```
nastyastavitskaya, line 10: **question:** have we decided on anything for the "special" size cases that do not match this type?

src/components/data-display/Collapse/Collapse.stories.tsx
```diff
@@ -276,7 +274,7 @@ export const ExampleCustomPanel: Story = {
         <Collapse
           bordered={false}
           defaultActiveKey={['1']}
-          expandIcon={({ isActive }) => <SplitIcon rotate={isActive ? 90 : 0} />}
+          expandIcon={({ isActive }) => <Icon name={isActive ? 'split' : 'add'} />}
```
nastyastavitskaya, line 277: **praise:** so much better! and easier to read 😍

src/components/general/Icon/Icon.tsx
jared-dickman: i was thinking the special overrides can be accomplished via css and the exposed iconId
[see here ](https://github.com/mParticle/aquarium/pull/182/files#diff-d62d4f275b8bde84ef2ed5dacef240befa00db19f57576fdbe28f18802b09879R90) for an example

src/components/data-display/Collapse/Collapse.stories.tsx
tibuurcio: second that!

src/components/general/Icon/Icon.tsx
```diff
@@ -1,11 +1,24 @@
...
+}
+
+export const Icon = (props: IIconProps) => {
+  const IconName = Icons[props.name]
+  const iconId = `icon-${props.name.toLowerCase()}`
+
+  let className = `${iconId} icon-size-${props.size ?? 'lg'}`
+
+  if (props.color) {
+    className += ` icon-color-${props.color.toLowerCase()}`
```
tibuurcio, line 20: **question:** why do we need the `.toLowerCase()` if colors are typed?

src/components/general/Icon/Icon.tsx
```diff
@@ -1,11 +1,24 @@
...
+  name: keyof typeof Icons
+  color?: IconColor
+  size?: IconSize
+}
+
+export const Icon = (props: IIconProps) => {
+  const IconName = Icons[props.name]
+  const iconId = `icon-${props.name.toLowerCase()}`
+
+  let className = `${iconId} icon-size-${props.size ?? 'lg'}`
```
tibuurcio, line 17: **thought:** I know we tend to use CSS but the usage of classes here makes it easier to customize in userland. I'm not sure that's the best approach for a component library. We can't prevent this fully but we should strive to at least make it harder. Given that shouldn't we use raw styles and a JS object to map sizes and colors instead?

src/components/navigation/GlobalNavigation/WorkspaceSelector/workspace-selector.css
```diff
@@ -86,14 +86,11 @@
...
-.workspaceSelector__activeIcon {
-  margin-right: var(--margin);
-  margin-left: var(--margin-xs);
-  width: var(--size-sm);
-  min-width: var(--size-sm);
-  height: var(--size-sm);
+  .icon-check {
+    margin-right: var(--margin);
+    margin-left: var(--margin-xs);
+  }
```
tibuurcio, line 93: Interesting, we are customizing internally and it's an example of how easy it is. I wonder if there's another way to achieve those, like implementing the margins from the other components instead of the icon 🤔

tibuurcio (commented):
Great work! Left some comments

src/components/general/Icon/Icon.tsx
tibuurcio: I think it wouldn't be a huge problem to support a `number` alongside the SizeProps just because we are not talking about a variant of the Icon component. Giving the users control of variants is a problem but here it's a property. Although we recommend consistent sizes, if a custom one is needed thats only a property changing and not a variant of the Icon.

src/components/general/Icon/Icon.tsx
jared-dickman: well the thought is we prefer to use css over js-in-css, because [the current ant js-in-css framework] is slow...
i had actually originally written this in js but re-wrote it in css..
i suppose its possible for user to create their own `.icon-size-lg` classes with higher specificity [or !importants] and ignore the ones exported from the aquarium
but.... we're not _really_ creating a *real* component library at the end of the day, its still a pretty custom library designed specifically for our use cases 🤔 🤔

src/components/general/Icon/Icon.tsx
jared-dickman: good point! this may have been from before i renamed the keys in the Icons object to be lowercase, and probs isnt needed anymore

src/components/general/Icon/Icon.tsx
jared-dickman: yes i think you are 🥸 

according to the design team [chatting with caro and sonia the other day], while we may want custom sizes, they will likely be very few and very far between. perhaps only a couple in the entire app

given that, from the perspective of a new dev, having a size as a first-class prop the same as the typed options doesnt make sense. someone may think using [any] numbers is _"fine"_, when in fact it is _"bad"_ and should not be easy to do

tibuurcio: Great point. I think I'm sold on the fact it should not be easy to do 👍🏼

src/components/navigation/GlobalNavigation/WorkspaceSelector/workspace-selector.css
gabyzif: I second this too (specially after reading [Gabe's](https://www.gabe.pizza/notes-on-component-libraries/) guide), but if you check the icons here for ex: 👍 💯 they have margins so the icons do not overlap.

src/components/general/Icon/Icon.tsx
tibuurcio: I think we should go the same route as the Icon size customization where we won't allow users to customize it easily.

For a small number of props like this we could just use React inline styles since those would be translated directly into HTML and not depend on any js-in-css lib. Performance would be determined by props changing and rerenders which would be the same as having classNames, but it would make it harder for userland to modify it.

src/components/general/Icon/icon.css
```diff
@@ -0,0 +1,75 @@
+.icon-size-xxl {
+  width: var(--size-xxl);
+  min-width: var(--size-xxl);
```
Seebiscuit, line 3: **question** Pardon my ignorance, but if we have `width` why do we need `min-width`?

src/components/general/Icon/Icon.tsx
tibuurcio: > we're not really creating a real component library at the end of the day, its still a pretty custom library designed specifically for our use cases

Agree, but then how do we decide on allowing customizations and not allowing them? I'm thinking since design is onboard with this we could always _not allow it_ easily and customize later if needed

src/assets/svg/chart-column.svg
tibuurcio, line 1: Is this padding correct? Seems bigger than the rest of the icons.

src/components/general/Icon/Icon.tsx
Seebiscuit: The sad fact is that, indeed, by providing the `iconId` in the css, now it is very easy to change ANY style attribute, just by creating a class ([Jared's link](https://github.com/mParticle/aquarium/pull/182/files#diff-d62d4f275b8bde84ef2ed5dacef240befa00db19f57576fdbe28f18802b09879R90)).

I want to double down on Jared's explanation, immediately above, about how _hard_ it should be to do _bad_ things. I suggest we remove the `iconId` class, and make it inconvenient (read nasty) to select the icon, and apply ANY style rules to it. 

However, because we sometimes do need escape hatches, we should add the `USAFE_` props to update size and color. We could even have a linter that throws a warning if it finds any `UNSAFE_` props in the changes. (It would be even nicer if would alert a codeowner.)

src/components/navigation/GlobalNavigation/WorkspaceSelector/workspace-selector.css
Seebiscuit: I third this. And it'd be necessary to have a component like @tibuurcio describes if we remove the `iconId` from the class name, like I suggested earlier.

src/components/general/Icon/Icon.tsx
Seebiscuit: Have we looked into CSS modules? That would solve the specificity problem and has none of the slowness of css-in-js.

src/components/general/Icon/Icon.tsx
jared-dickman: youve convinced me, i've removed the iconId className in the [latest commit](https://github.com/mParticle/aquarium/pull/182/commits/59820e77f6b85cedacd2977421139548ab3d3ab3)

not adding UNSAFE_ prefix yet, we can continue discussion and arrive at a decision

src/components/icons/README.md
```diff
@@ -1,4 +1,6 @@
 ﻿# Icons
 
+todo: needs update
```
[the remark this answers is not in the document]
jared-dickman: planning to update in this pr once the structure is finalized

src/components/general/Icon/icon.css
jared-dickman: yes its a great question, and i sadly dont have a great answer right now
they can both be [used together](https://css-tricks.com/almanac/properties/m/min-width/), though seems not as common to have them with the same value

i feel like this at one point solved an issue with the size, but cant point to a specific case. 
and can only say thats how the legacy design rules in indicative work `_tools.mixins.scss:17`

gabrielmarquesf (approved):
looks awesome to me!

src/assets/readme.md
```diff
@@ -0,0 +1,15 @@
+All icons are managed and approved by the design team. Do not add any assets that have not been reviewed by them.
+
+When adding a new svg:
+
+- ensure that the viewbox is a square [0 0 24 24]
```
tibuurcio, line 5: I think that changed
```suggestion
- ensure that the viewbox is a square [0 0 20 20]
```

src/assets/readme.md
```diff
@@ -0,0 +1,15 @@
...
+  - and that there is some internal padding so the image does not touch the edges of the box
+- ensure there are no hardcoded colors [unless otherwise desired]
+  - and all `fill` has the value `"currentColor"`.
+- ensure the svg should not set a `height` or `width` attribute
+
+Add the new asset to src/assets/svg, with a semantic name (one that describes the function and not the shape)
+Then add the icon as an import and export in src/components/icons/index.ts
+Then add the icon to src/constants/Icons.ts to allow usage in as the `<Icon/>` component `name` prop
```
tibuurcio, line 13: **nit**
```suggestion
Then add the icon to src/constants/Icons.ts to allow usage in as the `<Icon />` component `name` prop
```

src/components/general/Icon/Icon.tsx
```diff
@@ -1,11 +1,24 @@
...
+  color?: IconColor
+  size?: IconSize
+}
+
+export const Icon = (props: IIconProps) => {
+  const IconName = Icons[props.name]
+  const className = `icon-size-${props.size} icon-color-${props.color}`
+  const iconId = `icon-${props.name}`
+
+  return <IconName className={className} data-test={iconId} />
```
tibuurcio, line 18: **question** Are we sticking with classNames? I still think that's not the right approach for making it hard to do bad things.

Also about the data-test attribute I believe we want to allow that in userland but not sure about it directly from the library. What if a user wants to use a different data-test?

Seebiscuit: > Are we sticking with classNames? I still think that's not the right approach for making it hard to do bad things.

This is a worthwhile discussion.

While we are using classes, which are overridable, we are not exposing an easy way to target any icons. They don't have readily identifiable class names. So say you have an mP logo icon. It would render as:

```html
<svg ... class="icon-size-lg icon-color-default" ..> ...</svg>
```
To select it we'd have to do something like 

```css
svg.icon-size-lg.icon-color-default { 
   /* ...rules... */ 
}
```
and that's pretty darn general, isn't. I don't see anyone trusting this. Is this unsolvable? Ofc, not. We can always (and usually do) wrap the icon in something:

```js
//...
return <span className="'sneaky-wrapper'"><Icon ... /> </span>
```

and then we can easily target the icon:
```css
.sneaky-wrapper > svg { 
   /* ...rules... */ 
}
```
and the classes would be "easily" overriden. (Why quotes over "easily"? I'll touch on that below.)

So, using class names isn't totally foolproof, but it isn't ergonomic---it doesn't feel native---either. But, it could be argued that it's still too easy to override our expectations. 

So, to me, the argument centers around how we stop making it "easy" to override. Well, the most familiar thing I can think of is to upcharge our icon utility classes. What do I mean? Well, as much as I don't like it, we would add `!important` to the icon utility rules. 

The other alternative is what GT is probably thinking which is using straight styles. 

The takeaway, though, is that we should all agree that using straight class names is too easy. The second takeaway is that we should make a global decision about how we deal with this.

Good homework to answer the second takeaway: What would Antd do?

src/assets/readme.md
```diff
@@ -0,0 +1,15 @@
+All icons are managed and approved by the design team. Do not add any assets that have not been reviewed by them.
+
+When adding a new svg:
+
+- ensure that the viewbox is a square [0 0 24 24]
+  - and that there is some internal padding so the image does not touch the edges of the box
+- ensure there are no hardcoded colors [unless otherwise desired]
+  - and all `fill` has the value `"currentColor"`.
```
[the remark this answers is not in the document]
jared-dickman: thats where the "[unless otherwise desired]" part comes in. which yes is still pretty vague. giving specific exceptions or cases [like duo-tone] isnt a bad idea. ultimately it should be on design to provide us with icons that are build correctly [in terms of color and currentColor]

src/components/general/Icon/Icon.tsx
tibuurcio: I just realized I never proposed anything and I should have, sorry about that! Nice write summary of the problem gary 🙌🏼 

I think if we were to go with classes we could do what Antd does, which is to prefix it with something to identify it's coming from them. Something like `aquarium-icon-...`. This does not solve the exposing classes problem but at least make it explicit in userland that someone trying to override it would be overriding something specific to the Aquarium.

Now for not using classes, my proposal is to use raw styles indeed while we don't figure it out a better way to deal with styles that we do not want to expose. Reasoning being the only drawback I can think of is that it's CSS in JS but it's not any custom library. React should be fast on handling this.

The pros to using raw styles I can see are:
* Performance wise it's the same since re-renders will be caused by props changing and this would happen with classNames as well.
* It would be less code than using classNames
* And the biggest thing would be not exposing the classNames

The cons:
* CSS in JS

Does that make sense?

tibuurcio (approved):
[We've registered the decision](https://mparticle.slack.com/canvas/C06DS34GJ7N?focus_section_id=temp:C:OEVdfb37810e7024947aec8febc8) to go with CSS for now to ship this and keep discussing the approach on custom styles we will be using in the Aquarium

src/components/general/Icon/Icon.tsx
jared-dickman: from the Aquarium Decision Log, we agreed to maintain the current interface of the component and release the breaking change. We can move from css to js in a subsequent pr if deemed valuable or necessary. 

Another con of the inline styles cssinjs approach is flexibility to access pseudo-classes and pseudo-elements

```
 4/15/24: Today we decided to use css over css-in-js to customize the components mainly because it was very difficult to find the correct designToken to update the component and css has better performance than css in js
```

src/components/general/Icon/icon.css
```diff
@@ -0,0 +1,83 @@
...
+
+.icon-color-info {
+  color: var(--color-info);
+}
+
+.icon-color-text {
+  color: var(--color-text);
+}
+
+.icon-color-default {
```
jared-dickman, line 73: @tibuurcio should we prefix this with `aqua`

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -85,7 +83,7 @@ export const GlobalNavigation = (props: IGlobalNavigationProps) => {
                   onClick={() => {
                     props.onMpHomeClick()
                   }}>
-                  <MpLogo className="globalNavigation__mpSvg" />
+                  <Icon name="mpLogo" size="lg" color="white" />
```
jared-dickman, line 86: thought - this is the kind of change we want to see!

gabyzif (approved):

PR #185, opened 2024-03-29
feat: Allows publishing from feature branchs

tibuurcio:
## Summary

- With the decision to use trunk based development, the `dev` branch doesn't make much sense anymore. This PR removes it and makes it so that it's possible to publish versions from feature branches.

We still need to investigate edge cases but we need this merged on `main` to do it.

## Testing Plan

- Running `npx semantic-release --dry-run`  is generating the following version: `1.10.1-release-process.1`.
- [Release works](https://github.com/mParticle/aquarium/actions/runs/8485401033/job/23250104110)
- [NPM version](https://www.npmjs.com/package/@mparticle/aquarium/v/1.11.0-release-process.1)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# [1.11.0-release-process.1](https://github.com/mParticle/aquarium/compare/v1.10.0...v1.11.0-release-process.1) (2024-03-29)
+
+
+### Features
+
+* allows publishing from feature branches ([eb32f58](https://github.com/mParticle/aquarium/commit/eb32f58e8b616c27c51a081e56e4e8f8d751a171))
+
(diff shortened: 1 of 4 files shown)


jared-dickman:
i like this a lot!

only thought is, will the semantic branch name check be happy?

anl it does seem that since the last commit is from the automation bot, none of the check are [or will ever] run...

tibuurcio:
> i like this a lot!
> 
> only thought is, will the semantic branch name check be happy?
> 
> anl it does seem that since the last commit is from the automation bot, none of the check are [or will ever] run...

I think so because it only checks for startsWith and our branch name is the same, it's only the version on npm that changes.

Just saw that I called the branch feature/ not feat/ 🤦🏼 Have a feeling this is not the last time it will happen

tibuurcio:
Had to create another one with the new branch name #186

PR #186, opened 2024-03-29
feat: Allows publishing npm versions from feature branches

tibuurcio:
## Summary

- With the decision to use trunk based development, the `dev` branch doesn't make much sense anymore. This PR removes it and makes it so that it's possible to publish versions from feature branches.

We still need to investigate edge cases but we need this merged on `main` to do it.

## Testing Plan

- Running `npx semantic-release --dry-run`  is generating the following version: `1.10.1-release-process.1`.
- [Release works](https://github.com/mParticle/aquarium/actions/runs/8485401033/job/23250104110)
- [NPM version](https://www.npmjs.com/package/@mparticle/aquarium/v/1.11.0-release-process.1)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mparticle/aquarium",
-  "version": "1.10.0",
+  "version": "1.11.0-release-process.1",
   "description": "mParticle Component Library",
   "license": "Apache-2.0",
   "keywords": [
--- a/release.config.cjs
+++ b/release.config.cjs
@@ -1,6 +1,6 @@
 module.exports = {
   branches: ['main', {
-    name: 'feature/*',
+    name: 'feat/*',
     channel: 'feature',
     prerelease: "${name.split('/').slice(1).join('-').toLowerCase()}"
   }],
--- a/release.config.cjs
+++ b/release.config.cjs
@@ -1,5 +1,9 @@
 module.exports = {
-  branches: ['main', { name: 'dev', channel: 'dev', prerelease: 'dev' }],
+  branches: ['main', { name: 'dev', channel: 'dev', prerelease: 'dev' }, {
+    name: 'feature/*',
+    channel: 'channel-${name}',
+    prerelease: "prerelease-${name.split('/').join('-')}"
+  }],
   tagFormat: 'v${version}',
   repositoryUrl: 'https://github.com/mParticle/aquarium',
   plugins: [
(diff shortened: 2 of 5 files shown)


tibuurcio:
@jared-dickman yeah seems like what you mentioned is happening due to the last commit being from the bot.

jared-dickman:
this sounds similar to a problem that was happening in the ind frontend repo recently, Liam D helped solve it. he's super busy but if need be may be good to pick his brain on a solution

release.config.cjs
```diff
@@ -1,5 +1,9 @@
 module.exports = {
-  branches: ['main', { name: 'dev', channel: 'dev', prerelease: 'dev' }],
+  branches: ['main', {
+    name: 'feature/*',
```
jared-dickman, line 3: should this be `feat` instead of `feature`??

jared-dickman:
> @jared-dickman yeah seems like what you mentioned is happening due to the last commit being from the bot. We might have a way to make it commit first. Will test on monday.

seems this may be by design 
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

```When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN, with the exception of workflow_dispatch and repository_dispatch, will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's GITHUB_TOKEN, a new workflow will not run even when the repository contains a workflow configured to run when push events occur.```

File: release.config.cjs (hunk 1 above)
Comment (tibuurcio) [reply release.config.cjs:3]: yes!! thanks for the sharp eye as always!

**Review by jared-dickman — APPROVED** [review]

File: package.json [hunk 2]
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mparticle/aquarium",
-  "version": "1.10.0",
+  "version": "1.11.0-release-process.1",
```
tibuurcio, line 3: This will be updated as soon as we release a new version from main, but it makes me wonder if this should go into main? Probably not.

gabyzif (approved):

package.json
tibuurcio: It actually makes sense. The *last* version we will have once this reaches main is this one. And once we release a new one from main, that's gonna be updated. All good I think.

PR #193, opened 2024-04-03
fix: remove "@supports selector()" from css

nastyastavitskaya:
## Summary

- While updating the Cortex to the latest version of the package, the build was failing. After some investigation, the issue was found with `@supports selector()` syntax that was used for the global nav styling. postcss-loader package in Cortex is outdated and probably does not know how to process this syntax. Updating packages in Cortex is complex and requires some time. As an easy fix, removing the query is more suitable, since it keeps the same logic in place and should resolve the compilation errors.

## Testing Plan

- Run global navigation stories. Confirm the scroll bar is using an updated style.

### DIFF
--- a/src/components/navigation/GlobalNavigation/global-navigation.css
+++ b/src/components/navigation/GlobalNavigation/global-navigation.css
@@ -34,21 +34,19 @@
 }
 
 /* Otherwise, use `::-webkit-scrollbar-*` pseudo-elements */
-@supports selector(::-webkit-scrollbar) {
-  .globalNavigation::-webkit-scrollbar {
-    /* stylelint-disable-next-line property-no-vendor-prefix */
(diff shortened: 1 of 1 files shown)


tibuurcio (approved):
LGTM although we should have a ticket to investigate this in Cortex codebase so we could bring this back in the future.

It shouldn't be the lib's responsability to take care of this but we are not able to prioritize this from the Predictions suite side right now :/

nastyastavitskaya:
> LGTM although we should have a ticket to investigate this in Cortex codebase so we could bring this back in the future.
> 
> It shouldn't be the lib's responsability to take care of this but we are not able to prioritize this from the Predictions suite side right now :/

100%. ticket to investigate the issue in cortex codebase: https://mparticle-eng.atlassian.net/browse/UNI-439

PR #200, opened 2024-04-09
feat: adding QueryItem components

brewdente:
## Instructions

## Summary

1. Query Item Action component - This represents an action done in our query builders - something like "Add criteria" or "Add event attribute"
2. Query Item Qualifier component - This represents a selector of qualifiers used in building a query expression - for example, in "X attribute is <greater than/less than/equal to> Y value" the angle-bracketed list is the qualifier of that expression.

## Testing Plan

- Closes https://go.mparticle.com/work/UI3DM-1067 
- Closes https://go.mparticle.com/work/UI3DM-1068

## Notes
- I (Zoe Hayes/ZoeAstra) am taking this over from Kevin due to him taking a leave of absence.
- Nicolas Merlis did commit 8f0b7bd, but was unable to commit it himself due to access issues - so I did so on his behalf.


src/utils/utils.spec.ts
```diff
@@ -29,9 +29,10 @@ describe('Testing utils', () => {
 
     it('it should return "Windows" when the user agent includes "Win"', () => {
       // arrange
-      global.navigator = {
+      const navigatorOverride: Partial<Navigator> = {
```
brewdente, line 32: **thought** all changes in this file fix lint issues

src/components/data-entry/QueryItem/query-item.css
```diff
@@ -0,0 +1,7 @@
+.query-item-qualifier__item-selected-icon {
+  margin-left: 5px;
```
jared-dickman, line 2: id expect this to use a multiple of 4 via a var like `margin-left: var(--margin-xxs)`

src/utils/utils.spec.ts
jared-dickman: how bizarre

src/components/data-entry/QueryItem/query-item.css
ZoeAstra: I believe it will. This is being reworked to use the new tokens/variables that we just got merged into this branch 🙂

design/GlobalToken.json
```diff
@@ -183,7 +183,7 @@
     "value": "#f8f6fb"
   },
   "colorBgLayout": {
-    "value": "#fff"
+    "value": "#ffffff"
```
ZoeAstra, line 186: **thought** I'm not entirely sure why this has changed as of this export/token generation round. Maybe it was changed in Figma, or the export plugin changed something? Not important, just kinda weird and something to maybe keep an eye on.

jared-dickman: probably an ide setting, perhaps on paste

src/components/data-entry/QueryItem/QueryItem.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+import Action from './Action'
+import { Qualifier } from './Qualifier'
+
+export const QueryItem = () => {
+  return <>DON'T USE THIS OR YOU WILL BE FIRED!</>
```
brewdente, line 6: **question** @ZoeAstra how were you able to commit this?  I get an error in the pre-commit hook

src/components/data-entry/QueryItem/Action.tsx
```diff
@@ -0,0 +1,24 @@
+import { AddIcon, Button } from 'src/components'
+import './query-item.css'
+
+export interface IActionProps {
```
brewdente, line 4: **question** does an Action do anything?  I'd expect a callback or something.

src/components/data-entry/QueryItem/QueryItem.tsx
ZoeAstra: 🤷‍♀️

tibuurcio: OS differences?

src/components/data-entry/QueryItem/Action.tsx
```diff
@@ -0,0 +1,24 @@
...
+  isPrimary?: boolean
+  isDisabled?: boolean
```
tibuurcio, line 6: I see those props map to some of the props from our Button component. Whenever I see a `boolean` prop I'm always wondering if it shouldn't be an enum. 

**suggestion** What about instead of using multiple booleans to infer props we used enums to make it better for future evolutions? Like `type: 'primary' | 'default' = 'default'`

tibuurcio: For disabled, I don't think we would need an enum, but also suggest we keep the same name from our components for consistency, so `disabled` instead of `isDisabled`.

src/components/data-entry/QueryItem/Qualifier.tsx
```diff
@@ -0,0 +1,33 @@
...
+      node.isSelected ? <CheckIcon className="query-item-qualifier__item-selected-icon" /> : null,
+    placement: 'bottomLeft',
+    popupMatchSelectWidth: false,
+    status: props.errorMessage ? 'error' : undefined,
+    suffixIcon: null,
+    variant: 'borderless',
+  }
+  return (
+    <>
+      <Select className="query-item query-item-qualifier__select" {...selectProps}></Select>
```
tibuurcio, line 29:
```suggestion
      <Select className="query-item query-item-qualifier__select" {...selectProps} />
```

tibuurcio: I feel this should be a suggestion/fix from formatting/linting 🤔

src/components/data-entry/QueryItem/QueryItem.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { QueryItem } from 'src/components'
+
+const meta: Meta<typeof QueryItem> = {
+  title: 'Aquarium/Data Entry/QueryItem',
+  component: QueryItem,
+  parameters: {
+    docs: {
+      description: {
+        component:
+          'DO NOT USE THIS OR YOU WILL BE FIRED! This is a parent component of this component group, and is not supposed to be used on its own.',
```
tibuurcio, line 11: 😆

src/components/data-entry/QueryItem/Qualifier.tsx
```diff
@@ -27,7 +28,7 @@ export const Qualifier = (props: IQueryItemQualifierProps) => {
   return (
     <>
       <Select className="query-item query-item-qualifier__select" {...selectProps}></Select>
-      {props.errorMessage && <span className="query-item__error">{props.errorMessage}</span>}
+      {props.errorMessage && <Typography.Text type="danger">{props.errorMessage}</Typography.Text>}
```
jared-dickman, line 31: @tibuurcio we should think through the when and how of using the Typography components, css utils, and bem classes

src/components/data-entry/QueryItem/Qualifier.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+// default
+const meta: Meta<typeof QueryItem.Qualifier> = {
+  title: 'Aquarium/Data Entry/QueryItem/Qualifier',
+  component: QueryItem.Qualifier,
+  args: {},
+}
+export default meta
+
+// stories
+export const Empty: Story = {}
```
jared-dickman, line 15: **Suggestion** conventions shows this would be Primary
```suggestion
export const Primary: Story = {}
```

src/styles/_variables.css
```diff
@@ -115,6 +115,31 @@
...
+  --mp-query-item-bg-color: #ffffff;
+  --mp-query-item-bg-color-hover: #f8f6fb;
+  --mp-query-item-bg-color-active: #ebe8f8;
+  --mp-query-item-bg-color-disabled: #faf9f8;
+  --mp-query-item-border-width-active: 0 0 2px 0;
+  --mp-query-item-shadow-focus: 0 0 0 2px rgba(54, 0, 209, 0.1);
+  --mp-query-item-color-disabled: #505249;
+  --mp-query-item-value-selector-font-weight: 600;
+  --mp-query-item-value-selector-color: #20007a;
+  --mp-query-item-padding: 4px;
```
jared-dickman, line 132: curious what only having a single padding property [and not margin] could mean

is less more? is using padding and margin interchangeably at this level ok?

src/utils/utils.spec.ts
tibuurcio: Addressed at #222

src/components/data-entry/QueryItem/Cascader.tsx
```diff
@@ -0,0 +1,172 @@
...
+  let inputClasses = `query-item`
+  if (isOpen) inputClasses += ' query-item--open'
+  if (selectedValue && selectedValue.length != 0) inputClasses += ' query-item--selected'
+  if (props.errorMessage) inputClasses += ' query-item--error'
+
+  return (
+    <>
+      <BaseCascader {...baseProps}>
+        <Input
+          readOnly={true}
```
nrmerlis, line 125:
```suggestion
          readOnly
```

src/components/data-entry/QueryItem/Cascader.tsx
```diff
@@ -0,0 +1,172 @@
...
+          placeholder={props.placeholder}
+          status={props.errorMessage ? 'error' : undefined}
+          className={inputClasses}
+          value={selectedDisplayValue ?? selectedValue.slice(-1)}
```
nrmerlis, line 129: Don't you think putting this as a `const` or `fn` would be more declarative?

src/components/data-entry/QueryItem/Cascader.tsx
```diff
@@ -0,0 +1,172 @@
...
+          prefix={
+            props.icon ? (
+              <Icon name={props.icon} size="ms" color="primary" />
+            ) : (
+              <Icon name="empty" size="ms" color="primary" />
+            )
```
nrmerlis, line 135: `getIcon()` ?

src/components/data-entry/QueryItem/Cascader.tsx
```diff
@@ -0,0 +1,183 @@
...
+      setIsOpen(value)
+      if (value) setSearchValue('')
+    },
+  }
+
+  let inputClasses = `query-item`
+  if (isOpen) inputClasses += ' query-item--open'
+  if (selectedValue && selectedValue.length != 0) inputClasses += ' query-item--selected'
+  if (props.errorMessage) inputClasses += ' query-item--error'
+  const displayValue = selectedDisplayValue ?? selectedValue.slice(-1)
```
tibuurcio, line 120: Reusable case for our `last` util function from a previous suggestion 😃

src/components/data-entry/QueryItem/Cascader.tsx
```diff
@@ -0,0 +1,183 @@
...
+            lowerSource.indexOf(valueToHighlight),
+            lowerSource.indexOf(valueToHighlight) + valueToHighlight.length,
+          )}
+        </span>
+        {highlightMatches(
+          source.slice(lowerSource.indexOf(valueToHighlight) + valueToHighlight.length),
+          valueToHighlight,
+        )}
+      </>
+    )
```
tibuurcio, line 156: **suggestion** I would suggest us to avoid using ternaries on long line components like that cause it hurts reading experience. Ternaries are great for small React stuff but I feel if statements and early returns here would make it better. Let me know if you feel the same.

```suggestion
    if (lowerSource.indexOf(valueToHighlight) === -1) {
      return <>{source}</>
    }
    
    // return the rest of the component
   return ...
```

src/components/data-entry/QueryItem/Cascader.tsx
```diff
@@ -0,0 +1,183 @@
...
+          className={inputClasses}
+          value={displayValue}
+          prefix={getIcon()}
+        />
+      </BaseCascader>
+      {props.errorMessage && <Typography.Text type={'danger'}>{props.errorMessage}</Typography.Text>}
+    </>
+  )
+
+  function highlightMatches(source: string, valueToHighlight: string): ReactNode {
```
tibuurcio, line 138: I was going to suggest we change that to a capitalized function so that React identifies it as a component but I'm not 100% sure about the suggestion so let's not change it right now. Still it's interesting to [follow the discussion](https://mparticle.slack.com/archives/G3BBD5SE6/p1715623788913859) around this if you want :)

src/components/data-entry/QueryItem/Text.tsx
```diff
@@ -0,0 +1,12 @@
+import { Typography } from 'src/components/general/Typography/Typography'
+
+export interface ITextProps {
+  disabled?: boolean
+  text: string
+}
+
+export const Text = ({ disabled = false, text }: ITextProps) => {
+  return <Typography.Text disabled={disabled}>{text}</Typography.Text>;
+}
```
tibuurcio, line 10: **question** Why do we need this `<Text>` component given we are just passing the same props to `Typography.Text`?

tibuurcio (commented):
Wonderful work @ZoeAstra and team. Please treat all my comments as suggestions since I know you're on a tight schedule. If you can't get to something just let me know in the comments and I can create follow up tech debt **if needed**.

Also saw some instances of formatting differences. Not commenting on that cause prettier should take care of it. Is it running on commit?

src/components/data-entry/QueryItem/Cascader.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const [searchValue, setSearchValue] = useState('')
+  const [selectedValue, setSelectedValue] = useState<(number | string)[]>(props.value ?? [])
+  const [selectedDisplayValue, setSelectedDisplayValue] = useState(
+    props.value ? (props.value.slice(-1)[0] as any).label : '',
+  )
+  const [isOpen, setIsOpen] = useState(false)
+
+  useEffect(() => {
+    setItems(props.options)
+  }, [props.options])
```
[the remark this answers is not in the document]
ZoeAstra: So I just read a lot on this, and I think I figured it out. At some point I copied this code from an Ant example, and they had the options re-declared as state for some reason. I don't know why. But I kept it, not realizing that with the prop supplying the initial state it would never re-render with new data from the prop if I didn't do something weird. So I did this, so that the state would register the prop changing and update, when I could have just used the prop. Maybe. 🤔 🤷‍♀️ 

Also I found some articles saying that props updating doesn't always trigger a re-render, so maybe I even ran into difficulties there that just confused me more. Online React resources almost always have another source just below them in the search results, with an opposite claim, but *both* have working examples, because though they claim to be about the same thing, they both missed something very subtle about react that completely changes the way things work. 😵‍💫 😆 😭

src/components/data-entry/QueryItem/Cascader.tsx
ZoeAstra: Interesting, i never thought about how React handles these sorts of small subcomponents being influenced by name. Definitely keep everyone up to date on what we decide 👍

src/components/data-entry/QueryItem/Text.tsx
ZoeAstra: We don't! I just never got around to refactoring this away after it was added and immediately deemed useless.

jared-dickman (approved):

src/components/data-entry/QueryItem/Cascader.tsx
jared-dickman: agree, def each condition of the ternary could be pulled out into its own named `renderPart` function

src/components/data-entry/QueryItem/Cascader.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  const onSearch = ({ target: { value: value } }: { target: { value: string } }) => {
+    if (debouncedLoadData) {
+      if (value.length > 3) {
+        if (transformOptionsToPaths(items, []).filter(path => filter(value, path)).length == 0) {
```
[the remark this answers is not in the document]
jared-dickman: more of a stylistic thing but i agree, one line is sufficient here if everything has a simple named var

src/components/data-entry/QueryItem/Cascader.tsx
jared-dickman: its not about name its about scope, if a "component" is declared inside another component or not 

im trying to find the article on [developerway](https://www.developerway.com), i can picture it

but basically if you move the component definition outside of the parent component, it wont rerender every time the parent rerenders

src/components/data-entry/QueryItem/Cascader.tsx
```diff
@@ -0,0 +1,183 @@
...
+    },
+    options: items,
+    onDropdownVisibleChange: value => {
+      setIsOpen(value)
+      if (value) setSearchValue('')
+    },
+  }
+
+  let inputClasses = `query-item`
+  if (isOpen) inputClasses += ' query-item--open'
```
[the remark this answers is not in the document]
ZoeAstra: I think that's a good idea when we actually have more than one state, but until then: YAGNI 😄

PR #207, opened 2024-04-16
feat: UPS code ported from Nancy

tibuurcio:
## Summary
- This PR ports the code from the User Preferences Services from Nancy into the Aquarium to be used by the platforms
- Code was updated to make sure tests are passing but there's still some leftover tech debt to be addressed due to eslint. I've created [a TD ticket for it](https://mparticle-eng.atlassian.net/browse/UNI-589).

## Testing Plan

- Closes https://go.mparticle.com/work/UNI-485


src/services/Cookies.ts
tibuurcio, line 1: We need a package for this otherwise we gotta test all this code

src/services/user-preferences.ts
tibuurcio, line 1: It's all in one file for now. I'm gonna break it soon :)

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+# [1.13.0-ups-utils.1](https://github.com/mParticle/aquarium/compare/v1.12.0...v1.13.0-ups-utils.1) (2024-05-14)
```
tibuurcio, line 1: This is here cause of the feature release version I did

src/services/Cookies.ts
gabrielmarquesf, line 1: I see it is under `services`, but it's more an `util` to me.

src/services/user-preferences/models/definitions/user-preference-scope-type.ts
```diff
@@ -0,0 +1,6 @@
+export enum UserPreferenceScopeType {
+  Global,
+  LevelOneScope,
+  LevelTwoScope,
+  LevelThreeScope,
```
gabrielmarquesf, line 5: suggestion: (maybe it's too late to suggest it 😄) instead of level one|two|three, could it be more meaningful? like Global|WS|Acct|Org (that is what I suppose it would actually map to)

gabrielmarquesf (approved):
lgtm, though I haven't thoroughly reviewed the UPS implementation .
I also left two suggestions.

src/services/user-preferences/models/definitions/user-preference-scope-type.ts
tibuurcio: It's a good point if we think about unification in the future, but each platform currently has it's own way of defining those levels. Indicative has accounts and workspaces (no org) and cortex only has accounts, so yeah don't think we will be able to change it right now 🥲

src/services/Cookies.ts
tibuurcio: Makes sense 👍🏼

gabrielmarquesf (approved):

PR #208, opened 2024-04-17
feat: adding QueryItem.ValueSelector.InputText 

nrmerlis:
## Instructions

## Summary

QueryItem -> ValueSelector -> InputText component has been created.

Default:

image

Hover:

image

Active:

image

Focus:

image

Disabled:

image

Error:

image

## Testing Plan

Testested on StoryBook.

- Closes https://go.mparticle.com/work/UI3DM-1069


src/components/data-entry/QueryItem/TextInput.stories.tsx
```diff
@@ -0,0 +1,26 @@
+import { type Meta, type StoryObj } from '@storybook/react'
+import { QueryItem } from 'src/components'
+
+const meta: Meta<typeof QueryItem.ValueSelector.TextInput> = {
+  title: 'Aquarium/Data Entry/QueryItem/ValueSelector/TextInput',
```
jared-dickman, line 5: thoughts about including `Query Item` in the name of the file, so it shows up in the story

src/components/data-entry/QueryItem/TextInput.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+  component: QueryItem.ValueSelector.TextInput,
+  parameters: {
```
jared-dickman, line 7: i feel like we dont use `parameters` for our stories, what does this do??

src/components/data-entry/QueryItem/TextInput.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { Typography } from 'src/components/general/Typography/Typography'
+
+interface ITextInputProps {
+  defaultValue?: string;
+  onChange: (value: string | number) => void;
+  disabled?: boolean;
+  errorMessage?: string;
+}
+
+export const TextInput = ({ defaultValue, onChange, disabled = false, errorMessage }: ITextInputProps) => {
```
jared-dickman, line 13: **suggestion** dont destructure props 🔢

src/components/data-entry/QueryItem/TextInput.tsx
```diff
@@ -0,0 +1,34 @@
...
+  
+  const _onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setValue(e.target.value)
+    onChange(e.target.value)
+  }
+
+  return (
+    <>
+      <Input
+        status={errorMessage && !disabled ? 'error' : undefined}
```
jared-dickman, line 24: **suggestion** `extract variable` for the ternary [really a binary] to keep the attributes clean

src/components/data-entry/QueryItem/query-item.css
```diff
@@ -1,140 +1,155 @@
...
 .query-item-qualifier__select.ant-select-disabled {
-  color: var(--color-text-disabled);
-  background-color: var(--mp-query-item-bg-color-disabled);
+    color: var(--color-text-disabled);
+    background-color: var(--mp-query-item-bg-color-disabled);
 
-  .ant-select-selection-item {
-    border-bottom-color: var(--mp-query-item-border-color-disabled);
-  }
+    .ant-select-selection-item {
```
jared-dickman, line 127: this spacing isnt correct via the aquarium stylelint rules..

src/components/data-entry/QueryItem/TextInput.tsx
```diff
@@ -0,0 +1,34 @@
...
+        disabled={disabled}
+        className={'query-item query-item-input-text'}
```
jared-dickman, line 26: hmm this doesnt seem correct via bem, have you seen the css [style guide](https://mparticle.atlassian.net/wiki/spaces/FRON/pages/3880452097/Indicative+Style+Guide)? a lot of content in there

but basically, nothing should have two block or two element names. here `query-item` and `query-item-input-text` are both blocks in bem


something like `query-item query-item--input-text` would be ok. also ok would be only `query-item__input-text`

the question of which to pick, is _is this a special version of a query item, or a part of a query item_?

src/components/data-entry/QueryItem/TextInput.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+  },
+  args: {},
+}
+
+export default meta;
+
+type Story = StoryObj<typeof QueryItem.ValueSelector.TextInput>
+
+export const Primary: Story = {
+  args: {
```
jared-dickman, line 23: Primary stories dont usually have any args. this can be moved into the args on line 15 inside meta, so that it gets applied to every future story

src/components/data-entry/QueryItem/TextInput.stories.tsx
ZoeAstra: It lets you put a description in Storybook, so when browsing it, you can see what the component's intended purpose is. For example, this is from the WIP feat/query-item main branch: 
image

src/components/data-entry/QueryItem/TextInput.stories.tsx
mpartmer: What do you mean by "thoughts about including Query Item in the name of the file"?

Writing the tittle like Aquarium/Data Entry/QueryItem/ValueSelector/Query Item Text Input?

mpartmer: > I've thought about this. We had it that way at first, but it got wordy and repetitive, like QueryItem.QueryItemValueSelector.QueryItemTextInput, with IQueryItemTextInputProps. Unnecessary prefixing should be avoided, IMO.

> so if the path is paid attention to, it should be obvious which is which.

I agree with you @ZoeAstra . In my opinion it is not necessary at all

src/components/data-entry/QueryItem/TextInput.tsx
ZoeAstra: I was trying to see if all of the styling could be covered by one single `query-item` block-level class, but it's gotten a bit away from me since others have started chipping in. This will end up being refactored in the PR/branch that this one is being merged into.

I am also not sure if Nico has been introduced to BEM yet? May be worth giving another talk on it some time since we've had several newbies since the last one.

mpartmer: What is BEM????? :thinking_face: :thinking_face:

mpartmer: And having one single `query-item` block-level class is not possible because when we change the component status to an 'error' one, the ant-input-status-error class is applied, and it overrides, again, many previous overridden styles.

ZoeAstra: Oh, I also realized I never answered your question, Jared. The intention was to treat all of these as special instances/variants of one master QueryItem component. So using one class with modifiers, and the occasional specific element-level class, makes sense thematically.

src/components/data-entry/QueryItem/TextInput.stories.tsx
jared-dickman: thats great! amazing i havent heard of this before..

src/components/data-entry/QueryItem/TextInput.tsx
jared-dickman: lets chat about this before merging it in, i def have thoughts. and since this is the beginning of the Query Item && Query Builder legacy, i think its important to do it carefully and thoughtfully

ZoeAstra (approved):
Looks good. Couple things that will need to be nudged just a bit (through no fault of anyones, I just learned certain things have to be done a certain way when using custom controls in a Form... I had to make changes to mine and Kevin's already) when it's in the feat/query-items branch, but it would make more sense to do so there.

PR #209, opened 2024-04-17
feat: allow the Menu Item to include a Button at the bottom

gabyzif:
…lick handlers

## Instructions

## Summary

This PR implements an approach to conditionally render a button within the submenu of the navigation menu. The button is rendered only if at least one child item within the submenu has an onClick handler. This approach enhances the user interface of the navigation menu, providing dynamic functionality based on the presence of event handlers.

Screenshot 2024-04-17 at 4 15 23 PM

## Testing Plan

- Closes https://go.mparticle.com/work/UNI-427


gabyzif:
Note: If we decide to go with this aproach, I'll change the name of `WorkspaceSignout` f

PR #211, opened 2024-04-17
feat: add button to global nav menu

gabyzif:
…lick handlers

## Instructions

## Summary

This PR implements an approach to conditionally render a button within the submenu of the navigation menu. The button is rendered only if at least one child item within the submenu has an onClick handler. This approach enhances the user interface of the navigation menu, providing dynamic functionality based on the presence of event handlers.

Screenshot 2024-04-17 at 4 15 23 PM

## Testing Plan

- Closes https://go.mparticle.com/work/UNI-427


src/components/navigation/GlobalNavigation/NavigationList.tsx
```diff
@@ -35,13 +36,19 @@ function generateMenuItem(item: IGlobalNavigationItem, i: number) {
 
   if (item.type === 'menu') {
     children.push(
-      ...item.children.map((linkItem, j) => ({
+      ...item.children.filter(linkItem => !linkItem.onClick).map((linkItem, j) => ({
```
jared-dickman, line 39: not clear why we're filtering menu items with onClick
could we pull the thing we map on into a var with a name??

src/components/navigation/GlobalNavigation/NavigationList.tsx
```diff
@@ -35,13 +36,19 @@ function generateMenuItem(item: IGlobalNavigationItem, i: number) {
...
         label: buildLinkFromHrefOptions(linkItem.label, linkItem.hrefOptions),
       })),
     )
+    if (item.children.some(child => child.onClick)) {
+      children.push({
+        key: 'submenu-button',
+        label: <WorkspaceSignout signoutOptions={{ label:"test", onSignout: () => {} }} />,
+      });
+    }
   }
```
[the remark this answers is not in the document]
jared-dickman: agree using onClick as a boolean is not recommended. and for the nav, i feel like we should give a more rigid [but configurable] template than other exported components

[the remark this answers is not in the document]
gabyzif: I agree we should go more rigid and that the onclick is not good, that being said do you guys want to do the change it now or in the future?

src/components/navigation/GlobalNavigation/NavigationList.tsx
```diff
@@ -34,14 +35,24 @@ function generateMenuItem(item: IGlobalNavigationItem, i: number) {
...
+      ...linkItem,
+      expandIcon: null,
+      key: `${String(linkItem.label)}${j}`,
+      label: buildLinkFromHrefOptions(linkItem.label, linkItem.hrefOptions),
+    }));
+
+    const regularItems = submenuItems.filter(linkItem => linkItem.type !== 'button');
+    const actionItems = submenuItems.filter(linkItem => linkItem.type === 'button');
+
+    children.push(...regularItems);
```
jared-dickman, line 48: is this to ensure the buttons always go at the bottom?

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelectorContent.tsx
```diff
@@ -12,11 +12,11 @@ type WorkspaceSelectorContentProps = {
   inputRef: RefObject<InputRef>
   hasNoResults: boolean
   menuItems: IWorkspaceSelectorDisplayItem[]
-  signoutOptions?: IWorkspaceSelectorProps['signoutOptions']
```
jared-dickman, line 15: i like making it more generic but note this will be a breaking change in the platforms 🙁

src/components/navigation/GlobalNavigation/NavigationList.tsx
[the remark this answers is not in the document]
jared-dickman: yes looks like you already started in the last commit, i like abstracting the signout button for this usecase!

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
```diff
@@ -98,6 +98,9 @@ const defaultManagement: IGlobalNavigationItem[] = [
       { label: 'option 1', hrefOptions: { href: '/' } },
       { label: 'option 2', hrefOptions: { href: '/' } },
       { label: 'option 3', hrefOptions: { href: '/' } },
+      { label:"button", type: 'button', buttonOptions: { label:"button", onClick:() => alert("go")} },
+      { label:"go", type: 'button', buttonOptions: { label:"go", href:"/", target:"_blank" }}  
+
```
gabyzif, line 103: this is to show that we can add multiple buttons, with onclick or href.

jared-dickman:
youll need to update the name to start with `feat: ` to pass `Reusable Workflows / Check PR for semantic title`

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
```diff
@@ -140,9 +143,13 @@ const meta: Meta<typeof GlobalNavigation> = {
     orgs: defaultOrgs,
     signoutOptions: {
       onSignout: () => {
-        alert('signing out!')
-      },
-    },
+        alert('signing out old!')
```
jared-dickman, line 146: is this still old? probably not necessary to add to the story

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -35,6 +35,10 @@ export interface IGlobalNavigationProps {
     label?: string
```
jared-dickman, line 35: can we remove this and `signoutOptions` now?

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
```diff
@@ -98,6 +98,9 @@ const defaultManagement: IGlobalNavigationItem[] = [
       { label: 'option 1', hrefOptions: { href: '/' } },
       { label: 'option 2', hrefOptions: { href: '/' } },
       { label: 'option 3', hrefOptions: { href: '/' } },
+      { label:"button", type: 'button', buttonOptions: { label:"button", onClick:() => alert("go")} },
+      { label:"go", type: 'button', buttonOptions: { label:"go", href:"/", target:"_blank" }}
```
tibuurcio, line 102: Looks like prettier is not running cause it should take care of that.

```suggestion
      { label:"button", type: 'button', buttonOptions: { label: "button", onClick: () => alert("go")} },
      { label:"go", type: 'button', buttonOptions: { label: "go", href: "/", target: "_blank" }}  
```

src/components/navigation/GlobalNavigation/NavigationButtonItem.tsx
```diff
@@ -0,0 +1,28 @@
...
+    label?: string;
+    onClick?: () => void;
+    options?: {
+        withoutContainer?: boolean;
+    };
+}
+
+export const NavigationButtonItem: React.FC<INavigationButtonItemProps> = ({ label = 'Sign Out of mParticle', onClick, options = {}, ...rest }) => {
+    const { withoutContainer = false } = options;
+console.log(onClick,"onclick")
```
tibuurcio, line 15:
```suggestion
```

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
```diff
@@ -28,6 +28,11 @@ import { PaddingXxs } from 'src/styles/style'
 export interface IWorkspaceSelectorProps {
   orgs: INavigationOrg[]
   avatarOptions?: IAvatarProps
+  navigationButtonItemOptions?: {
+    label?: string
```
tibuurcio, line 32: **question** What's the use case for a navigationButtonItem that doesn't have a label? I would assume it's always needed 🤔

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelectorContent.tsx
```diff
@@ -12,12 +12,17 @@ type WorkspaceSelectorContentProps = {
   inputRef: RefObject<InputRef>
   hasNoResults: boolean
   menuItems: IWorkspaceSelectorDisplayItem[]
+  navigationButtonItemOptions?: {
+    label?: string;
+    onClick: () => void;
+    withoutContainer?: boolean;
+  }; 
+  // @deprecated signoutOptions will be deprecated.Please avoid using it.
```
tibuurcio, line 20: Ideally this should show up on usage and whenever I saw this type of comments in codebases I've always seen it using the `/** */` syntax instead of `//`. Not sure if it makes any difference.

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelectorContent.tsx
```diff
@@ -38,7 +43,7 @@ export function WorkspaceSelectorContent(props: WorkspaceSelectorContentProps) {
 
       {hasNoResults ? <WorkspaceNoResults /> : <WorkspaceSelectorContentItems menuItems={menuItems} />}
 
-      <WorkspaceSignout signoutOptions={signoutOptions} />
+      <NavigationButtonItem label={props.signoutOptions?.label} onClick={props.signoutOptions?.onSignout}  {...props.navigationButtonItemOptions}/>
```
tibuurcio, line 46: **suggestion** I get the idea here but I think it would be best to avoid spread operator whenever possible. A user bypassing typescript could inject any sort of prop here we might not want them to be able to. [Like gabe told us](https://gabe.pizza/notes-on-component-libraries/#:~:text=Avoiding%20JSX%20spread%20on%20foreign%20data%20prevents%20weird%20bugs%20sometimes.) 😄 

I would suggest then we use an approach of getting the props outside the return. Something like:
`const label = props.navigationButtonItemOptions.label ?? props.signoutOptions`

I just thought of another thing. When I tested this approach in the past I saw there's a way to type this in order to not be able to use both at the same time. That would be interesting for this deprecation pattern we are introducing.

tibuurcio: [Take a look at those types](https://github.com/sindresorhus/type-fest/tree/main#:~:text=RequireAtLeastOne%20%2D%20Create%20a,disallows%20more%2C%20or)

tibuurcio (commented):
Nice work!

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
gabyzif: I left it here to check that it wasnt broken

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelectorContent.tsx
jared-dickman: do we have a philosophy around making things deprecated vs just removing them?

gabyzif: so we don't break the other platforms :)

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector.tsx
```diff
@@ -28,6 +28,11 @@ import { PaddingXxs } from 'src/styles/style'
...
+    onClick: () => void,
+    withoutContainer?: boolean
+  },
   signoutOptions?: {
```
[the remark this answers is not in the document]
gabyzif: no, the idea is leaving it here so we don't break all the platforms now

gabyzif:
Follow up: https://mparticle-eng.atlassian.net/browse/UNI-506

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelectorContent.tsx
tibuurcio: I think this should be our way to deprecating stuff, but we don't have anything figured it out in regards to updating the platforms. Ideally once a deprecation notice is introduced, linters would tell us and the platforms could be updated in their next package version bump. 

How does that sound?

jared-dickman: conceptually seems fine, but not sure how to manage that in practice. should we remove deprecated things when we bump into them in another feature? or have a list of things to be removed in the next release? tech debt tickets to remove the code after n sprints?

to be honest for things like this right now, im ok with just removing them and causing a breaking change, since we control the platforms and releases 
similar to the [Icons pr](https://github.com/mParticle/aquarium/pull/182) that was merged in, it breaks the existing Icons but does not deprecate them [only removes them]

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -63,17 +70,18 @@ export const GlobalNavigation = (props: IGlobalNavigationProps) => {
             {props.orgs ? (
               <WorkspaceSelector
                 orgs={props.orgs}
+                navigationButtonItemOptions={props.navigationButtonItemOptions}
                 signoutOptions={props.signoutOptions}
                 avatarOptions={props.avatarOptions}
               />
             ) : (
-              !!props.signoutOptions?.onSignout && (
+              !!props.navigationButtonItemOptions?.onClick()  || !!props.signoutOptions?.onSignout&& (
```
jared-dickman, line 78: seems strange we would include the `Sign Out` label and icon for any `navigationButtonItemOptions.onClick`. shouldnt it be somethingLike `navigationButtonItemOptions.label`?

jared-dickman: oh this condition is kind of an edge case too. it only shows if we are hiding the ws selector, which only happens currently in the ind demo

src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelectorContent.tsx
tibuurcio: Makes sense to me. As long as we don't introduce huge breaking changes it should be fine to iterate on the platforms for now since things a moving fast.

Also, while I was testing this PR in the CDP this new options confused me a bit cause I only wanted to add the bottom button and this doesn't relate to that, which makes me think we should split refactor PR's in the future.

For this one, let's go with it and remove the signoutOptions @gabyzif? Less tech debt for us to deal with.

src/components/navigation/GlobalNavigation/NavigationButtonItem.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export const NavigationButtonItem: React.FC<INavigationButtonItemProps> = ({ label = 'Sign Out of mParticle', onClick, options = {}, ...rest }) => {
+    const { withoutContainer = false } = options;
+    return (
+        <div className={!withoutContainer ? "workspaceSelector__signoutButtonContainer" : ""}>
+            <Button
+                className="workspaceSelector__signoutButton"
+                type="primary"
+                onClick={onClick}
+                {...rest}
```
jared-dickman, line 21: not a fan of this pattern, dont you think its better to be explicit with what we expect?

src/components/navigation/GlobalNavigation/NavigationList.tsx
```diff
@@ -32,16 +33,31 @@ function generateMenuItem(item: IGlobalNavigationItem, i: number) {
...
-      })),
-    )
+    const childrenWithExpandedIcons = item.children.map((child, index) => ({
+      ...child,
+      expandIcon: null,
+      key: `${child.label}${index}`,
+      label: buildLinkFromHrefOptions(child.label, child.hrefOptions),
+    }));
+
+    const regularItems = childrenWithExpandedIcons.filter(child => child.type !== 'button');
```
jared-dickman, line 44: could cut the number of loops down with a one forEach instead of two filters

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
jared-dickman: suggestion - either remove signoutOptions in this pr [since its trivial to update the interfaces in userland] or create a td item to get it done subsequently. 
either way may want tickets for each platform as well to release the interface updates

src/components/navigation/GlobalNavigation/GlobalNavigationItems.d.ts
```diff
@@ -19,10 +19,11 @@ export interface IGlobalNavigationMenu extends IBaseGlobalNavigationItem {
 }
 
 export interface IGlobalNavigationLink extends IBaseGlobalNavigationItem {
-  type?: 'link'
+  type?: 'link' | 'button'
   hrefOptions?: HrefOptions
   hideLabel?: boolean
-  onClick?: (e: MouseEvent) => void
+  onClick?: (e: MouseEvent) => void,
+  buttonOptions?:{ label: string, onClick?: () => void, href?: string, target?: 'self' | '_blank'}
```
jared-dickman, line 26: spacing 🙈

src/components/navigation/GlobalNavigation/NavigationButtonItem.tsx
```diff
@@ -0,0 +1,27 @@
...
+    onClick?: () => void;
+    options?: {
+        withoutContainer?: boolean;
+    };
+}
+
+export const NavigationButtonItem: React.FC<INavigationButtonItemProps> = ({ label = 'Sign Out of mParticle', onClick, options = {}, ...rest }) => {
+    const { withoutContainer = false } = options;
+    return (
+        <div className={!withoutContainer ? "workspaceSelector__signoutButtonContainer" : ""}>
```
jared-dickman, line 16: without container? should it be the parents responsibility to control the container rules? 
also seems funky to have the container defined a "signout" class

src/components/navigation/GlobalNavigation/NavigationButtonItem.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export interface INavigationButtonItemProps {
+    label?: string;
+    onClick?: () => void;
+    options?: {
+        withoutContainer?: boolean;
+    };
+}
+
+export const NavigationButtonItem: React.FC<INavigationButtonItemProps> = ({ label = 'Sign Out of mParticle', onClick, options = {}, ...rest }) => {
```
jared-dickman, line 13: probably this default label should be moved up a level? would be weird to have `Sign Out of mParticle` appear in multiple places

src/components/navigation/GlobalNavigation/NavigationList.tsx
```diff
@@ -32,16 +33,31 @@ function generateMenuItem(item: IGlobalNavigationItem, i: number) {
...
+    const actionItems = childrenWithExpandedIcons.filter(child => child.type === 'button');
+
+    children.push(...regularItems);
+
+    actionItems.forEach((actionItem, index) => {
+      children.push({
+        className: "navigation__button-item",
```
jared-dickman, line 51: **suggestion** bem syntax in the aquarium uses camelCase instead of a single dash for words in the same section 
`navigation__buttonItem`

src/components/navigation/GlobalNavigation/global-navigation.css
```diff
@@ -176,6 +176,16 @@
   color: var(--color-primary-text) !important;
 }
 
+.navigation__button-item {
```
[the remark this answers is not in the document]
jared-dickman: this change stands, even though i _just_ suggested to use `navigation__buttonItem` in a previous comment
thinking about it a bit more, it should def start with `globalNavigation` for consistency and clarity. thats the `block` level name
then the `element` level name is after the `__` and before the optional `--`. here button is also kind of vauge, and its never a button that lives right in the main-nav correct, its always inside a menu popout?
in that case, perhaps `menuButton` is better? still could be misinterpreted 

its almost like we want a sub element, like `globalNavigation__menu__button`, but thats "not allowed" and a solution is to create a new block. its ok to have several blocks working together
`navigationMenu__button` could be, if also renaming `globalNavigation__menu` to `navigationMenu`, or `navigationList` so it matches the [component](https://github.com/mParticle/aquarium/blob/cee86c3388448127747cf3155f49819195f38c51/src/components/navigation/GlobalNavigation/NavigationList.tsx#L21) name!

as always, lots of options. and a lot of thinking about naming which isnt great. its another point for a more automated css [injs] solution

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
```diff
@@ -74,6 +74,8 @@ const defaultManagement: IGlobalNavigationItem[] = [
       { label: 'option 1', hrefOptions: { href: '/' } },
       { label: 'option 2', hrefOptions: { href: '/' } },
       { label: 'option 3', hrefOptions: { href: '/' } },
+      { label: 'button', type: 'button', buttonOptions: { label: 'button', onClick: () => alert('go') } },
+      { label: 'go', type: 'button', buttonOptions: { label: 'go', href: '/', target: '_blank' } },
```
tibuurcio, line 78: Besides the CSS suggestion, last thing I can see before this gets ready is this `label` being duplicated. Let's keep only one?

src/components/navigation/GlobalNavigation/NavigationButtonItem.tsx
```diff
@@ -0,0 +1,31 @@
...
+export interface INavigationButtonItemProps {
+  label?: string
+  onClick?: () => void
+  withoutContainer?: boolean
+  href?: string
+  target?: '_blank' | '_self'
+}
+
+export const NavigationButtonItem: React.FC<INavigationButtonItemProps> = ({
+  label = 'Sign Out of mParticle',
```
jared-dickman, line 14: this seems funky.. its good to have all the platforms default to the same signout text, but having the default be in `NavigationButtonItem.label` seems a bit strong. surely most other NavigationButtonItems wont deal with signout

src/components/navigation/GlobalNavigation/NavigationButtonItem.tsx
```diff
@@ -0,0 +1,31 @@
...
+  onClick,
+  withoutContainer = false,
+  ...props
+}) => {
+  return (
+    <div className={!withoutContainer ? 'workspaceSelector__signoutButtonContainer' : ''}>
```
jared-dickman, line 20: this can be improved in a couple ways. should it be the responsibility of the parent to handle the container?
in which cases do we and do we not want the container applied?

src/components/navigation/GlobalNavigation/NavigationButtonItem.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <Button
+        className="workspaceSelector__signoutButton"
```
jared-dickman, line 22: **suggestion** lets de-couple signout from NavigationButtonItem

src/components/navigation/GlobalNavigation/NavigationButtonItem.tsx
gabyzif: 100%. 

In the tech debt task, this extra class that's for the singout button in the workspace selector should be in the parent : )

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
tibuurcio: Yeah I get options being together but think we should leave it outside for consistency with the other types

src/components/navigation/GlobalNavigation/NavigationButtonItem.tsx
gabyzif: 100%. Perfect task for the tech debt task

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
```diff
@@ -257,7 +260,7 @@ const mpManagement: IGlobalNavigationItem[] = [
     children: [
       { hrefOptions: { href: '/' }, label: 'Platform Settings' },
       { hrefOptions: { href: '/' }, label: 'User Settings' },
-      { hrefOptions: { href: '/' }, label: <Button type="primary">Switch to Legacy UI</Button> },
+      // { hrefOptions: { href: '/' }, label: <Button type="primary">Switch to Legacy UI</Button> },
```
tibuurcio, line 263: Let's keep that as an example?

```suggestion
      { hrefOptions: { href: '/' }, label: <Button type="primary">Switch to Legacy UI</Button> },
```

tibuurcio (approved):
Great work so far @gabyzif! I see we are creating a few tech debts here but also, this has been open for quite some time now. Approving this cause we've made good progress and also for us to move forward with M1.5, but let's make sure to add the things we need to work on in the [tech debt ticket](https://mparticle-eng.atlassian.net/browse/UNI-506). I've added mentions to both's @jared-dickman comments there to make sure we address them in the future.

I think that's the perfect PR to use as an example that we need better process at tech spec/change management for the Aquarium. We should also strive for decoupling  refactoring from feature PRs IMHO. 

I would love to test some tech spec process for something that I've been meaning to tackle ever since M1, but couldn't get to it.

> [Rethink tools/management props on the global navigation component UNI-339](https://mparticle-eng.atlassian.net/browse/UNI-339)

tibuurcio (approved):

tibuurcio (approved):

tibuurcio (approved):

PR #216, opened 2024-04-23
chore: Add statistic stories

alexs-mparticle:
## Instructions

## Summary

- Adds Example Stories for Statistics Component based on the examples provided by [Ant Design](https://ant.design/components/statistic)
- Includes visual update to ExampleStory.tsx component by adding a description field and a divider.

## Testing Plan

- Visually confirm that components display when running `npm run storybook` and that component controls properly update UI Examples.

- Closes https://go.mparticle.com/work/UNI-541


src/components/data-display/Statistic/Statistic.stories.tsx
```diff
@@ -1,14 +1,242 @@
 import { type Meta, type StoryObj } from '@storybook/react'
-import { Statistic } from 'src/components/data-display/Statistic/Statistic'
+import { Row, Col, Button } from 'antd'
+import { type valueType } from 'antd/es/statistic/utils'
```
tibuurcio, line 3: Can we import those from src? We tend to not use imports directly from Antd to try and simulate real world usage in stories 🙏🏼

src/components/data-display/Statistic/Statistic.stories.tsx
```diff
@@ -1,14 +1,242 @@
...
+              />
+            </Card>
+          </Col>
+        </Row>
+      </ExampleStory>
+    )
+  },
+}
+
+const deadline = Date.now() + 1000 * 60 * 60 * 24 * 2 + 1000 * 30 // Dayjs is also OK
```
tibuurcio, line 209: **suggestion** maybe just adding parenthesis to just represent better what are the groupings here? Even if it doesn't need it for precedence.

tibuurcio (approved):
Thanks for the contribution @alexs-mparticle! Left 2 small suggestions :)

src/components/data-display/Statistic/Statistic.stories.tsx
jared-dickman: should first export from Statistics 
```export { type AntValueType as valueType }```

jared-dickman (approved):
looks great!! wondeful cross-squad collaboration on this one alex 🎩

PR #221, opened 2024-05-01
feat: Adds SuitesReminder hook to get a consistent look across platforms for the reminder notification

tibuurcio:
## Summary

Ant suggests that we use hooks as the recommended way of showing notifications. This PR suggests an unified way of displaying the suites reminder notification so that we can re-use it in the platforms. It also ties it to the `GlobalNavigation` component cause this should only be used from that context.

There's some custom Ant CSS needed for this PR. Ant notifications are `absolute` positioned and they need to consider the GlobalNav width in the platforms, thus we need a custom `left` value. Ant gives us a way of customizing the `prefixCls` for the Notification DOM element and that's what's being used here.

## Testing Plan

- Check the latest GlobalNavigation story in Storybook

- Closes https://go.mparticle.com/work/UNI-468


src/hooks/SuitesReminder/suites-reminder.css
```diff
@@ -1,3 +1,3 @@
-﻿.ant-notification.ant-notification-bottomLeft {
-    left: var(--nav-width) !important;
+﻿.globalNavigation__suitesReminder.globalNavigation__suitesReminder-bottomLeft {
```
jared-dickman, line 1:
```suggestion
﻿.globalNavigation__suitesReminder.globalNavigation__suitesReminder--bottomLeft {
```

jared-dickman: or is this a name generated from ant?

src/hooks/SuitesReminder/useSuitesReminder.tsx
```diff
@@ -36,6 +34,12 @@ export const useSuitesReminder = (options: ISuitesReminderOptions): SuitesRemind
     message = DefaultMessage,
   } = options
 
+  const [api, contextHolder] = notification.useNotification({
```
jared-dickman, line 37: **suggestion** a more specific name than `api`, the [docs](https://ant.design/components/message) have `messageApi`, or `reminderMessageApi` incase there are several

src/hooks/SuitesReminder/suites-reminder.css
tibuurcio: I saw that but it's generated from Ant, yeah :(

src/hooks/SuitesReminder/useSuitesReminder.tsx
tibuurcio: Nice one. Maybe `notificationApi` since we're dealing with notifications? I'm suggesting that.

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
```diff
@@ -995,3 +996,44 @@ export const WorkspaceSearchWithNoResults: Meta<typeof GlobalNavigation> = {
     await userEvent.type(searchInput, '123{enter}')
   },
 }
+
+export const UseSuitesReminderHook: Story = {
+  render: props => {
+    const [openNotification, contextHolder] = useSuitesReminder({
```
jared-dickman, line 1002: nice!
would be a good place to write some tests too 😉

src/hooks/SuitesReminder/useSuitesReminder.tsx
```diff
@@ -0,0 +1,78 @@
+import 'src/styles/_variables.css'
+import './suites-reminder.css'
+
```
jared-dickman, line 3:
```suggestion
```

src/hooks/SuitesReminder/useSuitesReminder.tsx
```diff
@@ -0,0 +1,78 @@
...
+  title?: string
+  message?: string
+}
+
+type OpenNotificationFn = () => void
+type ContextHolder = ReactNode
+
+export type SuitesReminderHook = [OpenNotificationFn, ContextHolder]
+
+const DefaultReminderDuration = 4.5 // same as antd notification default duration
```
jared-dickman, line 22: **suggestion** add `as const` for all of these consts

gabyzif (approved):

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
```diff
@@ -995,3 +996,44 @@ export const WorkspaceSearchWithNoResults: Meta<typeof GlobalNavigation> = {
...
+      onRemindMeLater: () => {
+        alert('Remind me later')
+      },
+      onTakeMeThere: () => {
+        alert('Take me there')
+      },
+    })
+
+    return (
+      <Flex style={{ minHeight: 800, width: 600, border: '1px solid black' }} justify="space-between" vertical={false}>
```
gabyzif, line 1015: question:  shoundlnt' we use a design token for the border?

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
tibuurcio: Oh you're so right. I totally forget about it. Will try some 👍🏼

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
tibuurcio: I'm removing that, it was for debugging purposes and I missed removing it 😅

src/hooks/SuitesReminder/useSuitesReminder.tsx
```diff
@@ -0,0 +1,78 @@
...
+  } = options
+
+  const [notificationApi, contextHolder] = notification.useNotification({
+    prefixCls: 'globalNavigation__suitesReminder',
+    duration,
+    placement: 'bottomLeft',
+  })
+
+  const openNotification = (): void => {
+    const key = `notification-${Date.now()}`
```
alexs-mparticle, line 44: Might be a bit of a nit on my part, but I think using date conversion for a key might be overkill and might have a performance hit. Why not just create a simple increment function?

tibuurcio: Interesting. Haven't given much thought to it but am always concerned with performance. I think in this case we will be fine since it will never be the case of showing more than just a few notifications at a time. Also it's the way Ant suggests in their docs.

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
```diff
@@ -995,3 +996,65 @@ export const WorkspaceSearchWithNoResults: Meta<typeof GlobalNavigation> = {
     await userEvent.type(searchInput, '123{enter}')
   },
 }
+
+export const UseSuitesReminderHook: Story = {
+  play: async () => {
+    const alert = fn().mockImplementation(() => {})
+    global.alert = alert
```
tibuurcio, line 1003: Not sure that's the best way of mocking stuff in Storybook. We are still learning how to test in the Storybook context 😅

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
tibuurcio: Done!

gabyzif (approved):

gabyzif (approved):

PR #222, opened 2024-05-01
chore: eslint fixes

tibuurcio:
## Summary
ESLint fixes

## Testing Plan

### DIFF
--- a/src/components/general/Icon/Icon.stories.tsx
+++ b/src/components/general/Icon/Icon.stories.tsx
@@ -1,9 +1,7 @@
 import React from 'react'
-import { Meta, type StoryObj } from '@storybook/react'
+import { type Meta } from '@storybook/react'
 import { Icons } from 'src/constants/Icons'
-import { Center, Flex, Icon, IIconProps } from 'src/components'
-
-type Story = StoryObj<typeof Icon>
+import { Flex, Icon, type IIconProps } from 'src/components'
 
 export const IconTable: React.FC<IIconProps> = ({ color = 'black', size = 'lg' }) => (
   <div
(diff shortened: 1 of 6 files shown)


src/components/navigation/GlobalNavigation/NavigationList.tsx
```diff
@@ -37,7 +37,7 @@ function generateMenuItem(item: IGlobalNavigationItem, i: number) {
     const childrenWithExpandedIcons = item.children.map((child, index) => ({
       ...child,
       expandIcon: null,
-      key: `${child.label}${index}`,
+      key: `${String(child.label)}${index}`,
```
jared-dickman, line 40: what is this, java?!

src/components/general/Icon/Icon.stories.tsx
```diff
@@ -15,7 +13,7 @@ export const IconTable: React.FC<IIconProps> = ({ color = 'black', size = 'lg' }
       justifyItems: 'center',
     }}>
     {(Object.keys(Icons) as Array<keyof typeof Icons>).map(iconName => (
-      <Flex vertical align="center">
+      <Flex vertical align="center" key={iconName}>
```
jared-dickman, line 16: makes me think, should we add a `key` property to the Icon component for example

jared-dickman (approved):

src/components/navigation/GlobalNavigation/NavigationList.tsx
tibuurcio: 😆😆 I though about typechecking but it was sooo verbose that I just did this

src/components/general/Icon/Icon.stories.tsx
tibuurcio: Don't think so 🤔 
Icon should be a one off thing and whoever decides to use it as a list should define the key for it. Also, I think all React c omponents already accept a key right?

PR #233, opened 2024-05-15
fix: hide minimap on boolean

gabyzif:
## Summary

This PR hides the minimap until it's enabled

- Closes https://go.mparticle.com/work/UNI-560

Updated storybook: [here](https://65aec861fa4adbf0708aebd0-sdwdgsvkbd.chromatic.com/?path=/story/aquarium-navigation-globalnavigation--mp)

Screenshot 2024-05-15 at 10 01 56 PM

Screenshot 2024-05-15 at 10 02 22 PM


src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -37,13 +37,37 @@ export interface IGlobalNavigationProps {
     onClick: () => void
     withoutContainer?: boolean
   }
-  minimapOptions: {
-    href: string
-  }
+  minimapOptions?: { href: '/'; show?: true }
```
gabrielmarquesf, line 40: change type to boolean, and break lines

gabrielmarquesf:
```suggestion
  minimapOptions?: { 
     href: '/'; 
     show?: boolean 
  }
```

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -37,13 +37,37 @@ export interface IGlobalNavigationProps {
...
+      <Popover
+        content={() => <MiniMap href={props.minimapOptions?.href || '/'} />}
+        placement="rightBottom"
+        arrow={false}>
+        <Center
+          className="globalNavigation__mpHome"
+          onClick={() => {
+            props.onMpHomeClick()
+          }}>
+          <Icon name="mpLogo" size="lg" color="white" />
```
gabrielmarquesf, line 56: suggestion: wrap it into another function. It will avoid repeating it bellow and will make clear what `WillRenderMinimap` is doing

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -37,13 +37,37 @@ export interface IGlobalNavigationProps {
...
   }
-  minimapOptions: {
-    href: string
-  }
+  minimapOptions?: { href: '/'; show?: true }
 }
 export const GlobalNavWidth = 90 as const
 
 export const GlobalNavigation = (props: IGlobalNavigationProps) => {
+  const WillRenderMinimap = () =>
```
gabrielmarquesf, line 45: it's not nice to create components inside the render function, in more complex components, it can cause perf issues ([ref](https://www.developerway.com/posts/react-re-renders-guide#part3.1))
So, either move it outside the comp, or change it to be a function like `function willRenderMinimap() { return <></>; }`

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -37,13 +37,37 @@ export interface IGlobalNavigationProps {
...
+        content={() => <MiniMap href={props.minimapOptions?.href || '/'} />}
+        placement="rightBottom"
+        arrow={false}>
+        <Center
+          className="globalNavigation__mpHome"
+          onClick={() => {
+            props.onMpHomeClick()
+          }}>
+          <Icon name="mpLogo" size="lg" color="white" />
+        </Center>
```
tibuurcio, line 57: **suggestion** component naming feels weird to me. WillRender in this case can **not render** if show is false. I would suggest we split those and reflect what they actually mean. I'm thinking something like `<CenterContent>` and `<WithMinimap>`. What do you think? That's cause the center content is always the same and it's being duplicated here. We could try to compose it better.

What do you think?

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
gabrielmarquesf: suggestion: as it also can not render the minimap, and render the tooltip, rename it to something like `renderHomeButton`

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
Seebiscuit, line 40: **issue** These values are being incorrectly initialized in an interface definition,
```suggestion
  minimapOptions?: { 
      href: string; // since there are many valid links, like "http://some.link/...", "https://some.link/...", "/some.link/...", or even just "some.link" for relative routes
      show?: boolean
  }
```

Seebiscuit: **question** How is `IGlobalNavigationProps.minimapOptions.href` going to be used? I would think you'd need a list of href's for the miniman, or none at all? Who is responsible for setting links to the map? Is it the Aquarium (since they are always the same) or is the apps (since the links sometimes have to be relative and sometimes external)?

Seebiscuit: **suggestion** Consider offering to hide, since we consider the default to be show. 

```suggestion
  minimapOptions?: { 
      href: string;
      hide?: boolean;
  }
```

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
Seebiscuit, line 45: **praise** I like this extraction!

Seebiscuit: **suggestion** I prefer we turn this into a real function and define this towards the bottom. I never like having definitions of things at the top, because I'm forced to load my brain with things I may not have context for.

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
Seebiscuit: Just want to comment that we should try to name things with the most useful level of abstraction for humans. Do we care that we are centering content or is that a design detail? We probably care that we are:
1. Rendering the mP Home button
2. We are rendering a button with a Minimap
3. We are render a button with a Tooltip

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -37,10 +36,12 @@ export interface IGlobalNavigationProps {
     onClick: () => void
     withoutContainer?: boolean
   }
-  minimapOptions: {
-    href: string
+  minimapOptions?: {
+    goToOverviewHref: '/'
```
gabrielmarquesf, line 40:
```suggestion
    goToOverviewHref: string
```

src/components/navigation/GlobalNavigation/GlobalNavigationHelpers.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export function WithTooltip({ onClick }: { onClick: () => void }) {
+  return (
+    <Tooltip title="mParticle Overview" placement="right">
+      <div></div>
+      <CenterContent onClick={onClick} />
+    </Tooltip>
+  )
+}
```
gabrielmarquesf, line 31: 1 - these functions don't need to be exported, right?
2 - I would avoid using `with` prefix, just because this is a naming pattern for react HOC (high order components)

src/components/navigation/GlobalNavigation/GlobalNavigationHelpers.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export function renderHomeButton(props: IGlobalNavigationProps) {
```
gabrielmarquesf, line 33: as you moved this function outside the render function, you could keep it as a component because it will not cause a perf issue anymore.
```suggestion
export function HomeButton(props: IGlobalNavigationProps) {
```

src/components/navigation/GlobalNavigation/GlobalNavigationHelpers.tsx
```diff
@@ -0,0 +1,39 @@
...
+        <CenterContent onClick={onClick} />
+      </div>
+    </Popover>
+  )
+}
+
+export function WithTooltip({ onClick }: { onClick: () => void }) {
+  return (
+    <Tooltip title="mParticle Overview" placement="right">
+      <div></div>
```
gabrielmarquesf, line 27: do we need this `div`?

src/components/navigation/GlobalNavigation/GlobalNavigationHelpers.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <Center className="globalNavigation__mpHome" onClick={onClick}>
+      <Icon name="mpLogo" size="lg" color="white" />
+    </Center>
+  )
+}
+
+export function WithMinimap({ goToOverviewHref, onClick }: { goToOverviewHref: string; onClick: () => void }) {
+  return (
+    <Popover content={() => <MiniMap goToOverviewHref={goToOverviewHref} />} placement="rightBottom" arrow={false}>
+      <div>
```
gabrielmarquesf, line 17: do we need this `div`?

gabyzif: yes bc if not ant doesn't show the popup :/

src/components/navigation/GlobalNavigation/GlobalNavigationHelpers.tsx
gabrielmarquesf, line 1: Now we could rename this file to `HomeButton`

src/components/navigation/GlobalNavigation/GlobalNavigationHelpers.tsx
gabrielmarquesf: I see you removed it on your last commit. Curious to know what was the issue 😄

gabyzif: I was importing it from ant instead of our components

src/components/navigation/GlobalNavigation/GlobalNavigationHelpers.tsx
gabyzif: thanks :)

gabrielmarquesf (approved):

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -37,9 +37,11 @@ export interface IGlobalNavigationProps {
     withoutContainer?: boolean
   }
   minimapOptions: {
-    href: string
+    goToOverviewHref: string
```
tibuurcio, line 40: **suggestion** `goTo` feels like a callback/fn. Can we use `overviewHref` instead?

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -37,9 +37,11 @@ export interface IGlobalNavigationProps {
...
+    hide?: boolean
```
tibuurcio, line 41: **question** any particular reasion we've changed from `show` to `hide`? I tend to like the first one better just cause it will fit well with our flagging functions, like `show: isOnNewExperience()` instead of having to flip the logic it with `!`.

src/components/navigation/GlobalNavigation/HomeButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+  onClick: () => void
+}
+
+interface HomeButtonInterface {
+  minimapOptions: {
+    goToOverviewHref: string
+    hide?: boolean
+  }
+  onMpHomeClick: () => void
+}
```
tibuurcio, line 24: **opinion** I don't really like the `Interface` at the end of it. In mP's codebase there's a discussion about using `I` in front on interfaces but that's not decided yet. I believe those are props right? Could we do `Props` in the end to follow current Aquarium standards?

src/components/navigation/GlobalNavigation/HomeButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export const HomeButton: React.FC<HomeButtonInterface> = props => {
+  return props.minimapOptions?.hide ? (
+    <MinimapWithPopover
+      goToOverviewHref={props.minimapOptions?.goToOverviewHref || '/'}
+      onClick={props.onMpHomeClick}
+    />
+  ) : (
+    <TooltipWithButton onClick={props.onMpHomeClick} />
+  )
```
tibuurcio, line 51: It should be the opposite 😅 `!hide`. Although we might be changing that to show again and then it would be fixed :)

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
gabyzif: @tibuurcio this is the reason why I changed it from show to hide

gabrielmarquesf (approved):

PR #241, opened 2024-05-21
feat: add possibility to add background color to the suite icon logo

gabyzif:
This PR adds the possibility to add background color to the suite icon logo 

image

### DIFF
--- a/src/components/navigation/GlobalNavigation/GlobalNavigationItems.d.ts
+++ b/src/components/navigation/GlobalNavigation/GlobalNavigationItems.d.ts
@@ -11,6 +11,7 @@ export interface IBaseGlobalNavigationItem {
 
 export interface IGlobalNavigationLogo extends IBaseGlobalNavigationItem {
   onSuiteLogoClick: () => void
+  hasBackground?: boolean
 }
 
 export interface IGlobalNavigationMenu extends IBaseGlobalNavigationItem {
--- a/src/components/navigation/GlobalNavigation/SuiteLogo.tsx
+++ b/src/components/navigation/GlobalNavigation/SuiteLogo.tsx
@@ -1,11 +1,26 @@
-import { Center } from 'src/components'
+import { Center, Icon } from 'src/components'
 import { NavigationIcon } from 'src/components/navigation/GlobalNavigation/NavigationIcon'
 import { type IGlobalNavigationLogo } from 'src/components/navigation/GlobalNavigation/GlobalNavigationItems'
+import React from 'react'
+import { Icons } from 'src/constants/Icons'
 
 export function SuiteLogo(props: IGlobalNavigationLogo) {
   return (
-    <Center vertical className="globalNavigation__suiteLogo" onClick={props.onSuiteLogoClick}>
-      <NavigationIcon icon={props.icon} label="" hideLabel className="globalNavigation__icon--suiteLogo" />
+    <Center vertical className={`globalNavigation__suiteLogo`} onClick={props.onSuiteLogoClick}>
+      <NavigationIcon
+        icon={
+          typeof props.icon === 'string' && Object.keys(Icons).includes(props.icon) ? (
+            <Icon name={props.icon as keyof typeof Icons} color="brand" size="xxl" />
+          ) : (
+            props.icon
+          )
+        }
+        label=""
+        hideLabel
+        className={`${
+          props.hasBackground ? 'globalNavigation__suiteLogo-background' : 'globalNavigation__icon--suiteLogo'
+        }`}
+      />
       {props.label}
     </Center>
   )
--- a/src/components/navigation/GlobalNavigation/global-navigation.css
+++ b/src/components/navigation/GlobalNavigation/global-navigation.css
@@ -3,6 +3,7 @@
   --nav-item-width: 82px;
   --nav-width: 90px;
   --suite-logo-height: 82px;
+  --suite-logo-width: 82px;
   --color-scrollbar-handle: rgb(0 0 0 / 15%);
   --color-scrollbar-track: transparent;
   --size-scrollbar-handle: 6px;
@@ -240,10 +241,7 @@
   padding-bottom: var(--padding-xxs);
 }
 
-.globalNavigation__icon.globalNavigation__icon--suiteLogo svg {
-  width: 36px;
-  height: 36px;
-}
+
 
 .globalNavigation__menu .globalNavigation__iconLabel {
   margin-top: var(--margin-xxs);
@@ -285,6 +283,7 @@
   border: 1px solid var(--color-border, #c3aeff);
 }
 
+
 .globalNavigation__suiteLogo {
   min-height: var(--suite-logo-height);
   min-width: var(--nav-width);
@@ -293,6 +292,19 @@
   font-size: var(--font-size-sm);
   font-weight: 400;
   cursor: pointer;
+}
+
+.globalNavigation__suiteLogo-background {
+  background-color: var(--mp-brand-primary-2);
+  padding: 10px;
+  margin-bottom: 8px;
+  border-radius: 100%;
+}
+
+.globalNavigation__icon.globalNavigation__icon--suiteLogo svg {
+  width: 36px;
+  height: 36px;
+  
 }
 
 .globalNavigation__divider {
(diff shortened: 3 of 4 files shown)


src/components/navigation/GlobalNavigation/SuiteLogo.tsx
```diff
@@ -1,11 +1,28 @@
...
+import { Center, Icon } from 'src/components'
 import { NavigationIcon } from 'src/components/navigation/GlobalNavigation/NavigationIcon'
 import { type IGlobalNavigationLogo } from 'src/components/navigation/GlobalNavigation/GlobalNavigationItems'
+import React from 'react'
+import { Icons } from 'src/constants/Icons'
 
 export function SuiteLogo(props: IGlobalNavigationLogo) {
+  const classMap: { [key: string]: string } = {
+    default: 'globalNavigation__icon--suiteLogo',
+    'background-solid': 'globalNavigation__suiteLogo-background',
```
tibuurcio, line 10: I would say those follow the same structure if they represent different types of the same thing. So either both become `globalNavigation__icon--typeThingy` or `globalNavigation__suiteLogo--typeThingy`. Does that make sense?

src/components/navigation/GlobalNavigation/global-navigation.css
```diff
@@ -295,6 +290,19 @@
...
+.globalNavigation__suiteLogo-background {
+  background-color: var(--mp-brand-primary-2);
+  padding: 10px;
+  margin-bottom: 8px;
+  border-radius: 100%;
+}
+
+.globalNavigation__icon.globalNavigation__icon--suiteLogo svg {
+  width: 36px;
+  height: 36px;
```
tibuurcio, line 302: **question** Should those be variables?

src/components/navigation/GlobalNavigation/global-navigation.css
```diff
@@ -295,6 +290,19 @@
   cursor: pointer;
 }
 
+.globalNavigation__suiteLogo-background {
+  background-color: var(--mp-brand-primary-2);
+  padding: 10px;
+  margin-bottom: 8px;
```
tibuurcio, line 296: **question** should those be variables?

src/components/navigation/GlobalNavigation/SuiteLogo.tsx
```diff
@@ -1,11 +1,28 @@
...
-    <Center vertical className="globalNavigation__suiteLogo" onClick={props.onSuiteLogoClick}>
-      <NavigationIcon icon={props.icon} label="" hideLabel className="globalNavigation__icon--suiteLogo" />
+    <Center vertical className={`globalNavigation__suiteLogo`} onClick={props.onSuiteLogoClick}>
+      <NavigationIcon
+        icon={
+          typeof props.icon !== 'string' ? (
+            props.icon
+          ) : (
+            <Icon name={props.icon as keyof typeof Icons} color="brand" size="xxl" />
+          )
```
tibuurcio, line 20: **question** I'm confused why do we need that? 🤔 
Are we trying to support both a `string` and JSX is that it? That's for removing the Icon thing on mP's root page?

gabyzif: correct. IMO we should move towards the string with the pre-set size of the icon and color (specially for consistency purposes, since this is only used in the navbar)... but I don't wanna break everything (now)

gabyzif: right now, we are setting the icon size in the css, but since we have the new size variables we can just use them :)

src/components/navigation/GlobalNavigation/global-navigation.css
gabyzif: 100%

src/components/navigation/GlobalNavigation/SuiteLogo.tsx
gabyzif: 100%

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -50,7 +51,7 @@ export const GlobalNavigation = (props: IGlobalNavigationProps) => {
       <Layout.Sider className="globalNavigation__sider" width={GlobalNavWidth}>
         <Flex vertical justify="space-between" style={{ height: '100%' }}>
           <div>
-            <SuiteLogo {...props.logo} />
+            {!props.hideSuiteLogo && <SuiteLogo {...props.logo} />}
```
nastyastavitskaya, line 54: nit: personal preference, but maybe something like that with true by default? also aligned with minimap show pattern
```suggestion
            {props.showSuiteLogo && <SuiteLogo {...props.logo} />}
```

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -50,7 +51,7 @@ export const GlobalNavigation = (props: IGlobalNavigationProps) => {
...
             <div className="globalNavigation__divider" />
```
nastyastavitskaya, line 55: suggestion: I think this divider should also be hidden if suite logo is not showing. it has bottom margin, that's why we have this odd spacing

src/components/navigation/GlobalNavigation/global-navigation.css
```diff
@@ -295,6 +290,19 @@
   cursor: pointer;
 }
 
+.globalNavigation__suiteLogo-background {
+  background-color: var(--mp-brand-primary-2);
```
nastyastavitskaya, line 294: nit: hmmm, I'd prefer to see semantic names, but it's up to you (I'm not the CSS owner here 😝 )
```suggestion
  background-color: var(--color-primary-bg-hover);
```

gabyzif: Screenshot 2024-05-21 at 11 45 20 PM

jeje I used this variable to keep consistency w the icons :)

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
jared-dickman: yes good call out! often i find it simpler to read `if true` statements rather than `if !true`

src/components/navigation/GlobalNavigation/SuiteLogo.tsx
jared-dickman, line 10: **issue** bem naming here, the aquarium uses camelCase within a subugrouping [block/element/modifier]
```
'background-solid': 'globalNavigation__suiteLogoBackground',
```
or
```
'background-solid': 'globalNavigation__icon--suiteBackground',
```

src/components/navigation/GlobalNavigation/SuiteLogo.tsx
jared-dickman: huh, if we do go this route, i suggest flipping the if statement, so the true condition comes first 
`if icon == string, <icon/>, else props.icon`

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -30,6 +30,7 @@ export interface IGlobalNavigationProps {
   onSuiteLogoClick?: () => void
   onMpHomeClick: () => void
   hideMpHome?: boolean
+  showSuiteLogo?: boolean
```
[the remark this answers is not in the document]
gabyzif: I was telling nastya the same thing at the same time jejje

src/components/navigation/GlobalNavigation/GlobalNavigationItems.d.ts
```diff
@@ -11,7 +12,8 @@ export interface IBaseGlobalNavigationItem {
 
 export interface IGlobalNavigationLogo extends IBaseGlobalNavigationItem {
   onSuiteLogoClick: () => void
-  type?: 'default' | 'background-solid'
+  type?: 'default' | 'background-solid' | 'custom-size'
+  icon?: ReactElement | keyof typeof Icons
```
jared-dickman, line 16: can we not use `ReactNode` here? thats what we've been using for most types, do you [or Mr. A.I.] know the difference between the two?

gabyzif: I changed it to ReactElement bc ReactNode accepts string and I want to enforce to write only the strings that are allowed in the icon component :)

nastyastavitskaya (approved):
🥳 let's gooo

PR #243, opened 2024-05-23
fix: Align button content to the center

gabyzif:
## Whenever a button has an icon larger than the label's font size, it becomes misaligned

image

Currently, we are adding an icon through the icon prop. While version 5.17 (we are currently in 5.13) introduces iconPosition this prop only allows switching the icon between the left and right sides, not centering it.

Screenshot 2024-05-23 at 9 24 08 PM

Screenshot 2024-05-23 at 9 24 21 PM

Since we are adding it though props and not in the children, we can't use a Flex container to make the button aligned, and we can't let the children handle the alignment either. Here is a code snippet from ant's implementation on the icon/children. The rest of the code you can find it [here](https://github.com/ant-design/ant-design/blob/master/components/button/button.tsx#L261):

  ```
const genButtonContent = (iconComponent: React.ReactNode, kidsComponent: React.ReactNode) =>
    iconPosition === 'start' ? (
      <>
        {iconComponent}
        {kidsComponent}
      </>
    ) : (
      <>
        {kidsComponent}
        {iconComponent}
      </>
    );

```

As a workaround for this problem, I set up a class that makes the button flex and center aligned only when it has an icon

The button with the fix: 
Screenshot 2024-05-23 at 4 26 06 PM

### DIFF
--- a/src/components/general/Button/Button.tsx
+++ b/src/components/general/Button/Button.tsx
@@ -1,13 +1,16 @@
 import { Button as AntButton } from 'antd'
 import { type ButtonProps as AntButtonProps } from 'antd'
 import { ConfigProvider } from 'src/components/other/ConfigProvider/ConfigProvider'
+import './button.css'
 
 export interface IButtonProps extends AntButtonProps {}
 
 export const Button = (props: IButtonProps) => {
   return (
     <ConfigProvider>
-      <AntButton {...props}>{props.children}</AntButton>
+      <AntButton {...props} className="button">
+        {props.children}
+      </AntButton>
     </ConfigProvider>
   )
 }
--- a/src/components/general/Button/Button.tsx
+++ b/src/components/general/Button/Button.tsx
@@ -8,7 +8,7 @@ export interface IButtonProps extends AntButtonProps {}
 export const Button = (props: IButtonProps) => {
   return (
     <ConfigProvider>
-      <AntButton {...props} className="aquarium-button">
+      <AntButton {...props} className={props.icon ? 'aquarium-button--aligned-center' : ''}>
         {props.children}
       </AntButton>
     </ConfigProvider>
--- a/src/components/general/Button/Button.tsx
+++ b/src/components/general/Button/Button.tsx
@@ -8,7 +8,7 @@ export interface IButtonProps extends AntButtonProps {}
 export const Button = (props: IButtonProps) => {
   return (
     <ConfigProvider>
-      <AntButton {...props} className="button">
+      <AntButton {...props} className="aquarium-button">
         {props.children}
       </AntButton>
     </ConfigProvider>
--- a/src/components/general/Button/Button.tsx
+++ b/src/components/general/Button/Button.tsx
@@ -1,14 +1,13 @@
 import { Button as AntButton } from 'antd'
 import { type ButtonProps as AntButtonProps } from 'antd'
 import { ConfigProvider } from 'src/components/other/ConfigProvider/ConfigProvider'
-import './button.css'
 
 export interface IButtonProps extends AntButtonProps {}
 
 export const Button = (props: IButtonProps) => {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/components/general/Button/button.css [hunk 1]
```diff
@@ -0,0 +1,4 @@
+.button {
```
Seebiscuit, line 1: **question** I'm suprised that this is happening. My hunch is that maybe we've missed a Design Token. Have those been investigated?

src/components/general/Button/Button.tsx
```diff
@@ -1,13 +1,16 @@
...
 import { ConfigProvider } from 'src/components/other/ConfigProvider/ConfigProvider'
+import './button.css'
 
 export interface IButtonProps extends AntButtonProps {}
 
 export const Button = (props: IButtonProps) => {
   return (
     <ConfigProvider>
-      <AntButton {...props}>{props.children}</AntButton>
+      <AntButton {...props} className="button">
```
tibuurcio, line 11: Do we know if this class get's leaked into consumers? It worries me such a generic name. Also, if that's a problem on Antd we might want to patch their base button class instead 🤔

tibuurcio (commented):
I was obligated to left a comment by github

image

src/components/general/Button/Button.tsx
gabyzif: good catch :)

src/components/general/Button/button.css
gabyzif: The bottom button is a raw ant component without our theme on top nor the new class.

Screenshot 2024-05-23 at 5 01 23 PM

src/components/general/Button/Button.tsx
```diff
@@ -1,13 +1,16 @@
...
 import { ConfigProvider } from 'src/components/other/ConfigProvider/ConfigProvider'
+import './button.css'
 
 export interface IButtonProps extends AntButtonProps {}
 
 export const Button = (props: IButtonProps) => {
   return (
     <ConfigProvider>
-      <AntButton {...props}>{props.children}</AntButton>
+      <AntButton {...props} className="aquarium-button">
```
jared-dickman, line 11: hum, um, yea can we do this in a different way? why not use a utility class, or a `Flex` component?

also, why do we need every single button from the aquarium to always be flex? shouldnt it def be a props.children level concern? 

if this is fixed in 5.17, might it be better to find an update path instead of patch a hack with display?

src/components/general/Button/button.css
```diff
@@ -0,0 +1,4 @@
+.aquarium-button--aligned-center {
```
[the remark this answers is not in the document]
gabyzif: I was litterally changig it for the utility class

src/utils/utils.css
```diff
@@ -1,3 +1,11 @@
 .u-display-none {
   display: none !important;
 }
+
+.u-display-flex {
+  display: flex !important;
+}
+
+.u-align-items-center{
```
jared-dickman, line 9:
```suggestion
.u-align-items-center {
```

tibuurcio (approved):
Good job on the investigation, evolution of context and alignment seeking in this PR :)

src/components/general/Button/Button.tsx
```diff
@@ -7,7 +7,9 @@ export interface IButtonProps extends AntButtonProps {}
 export const Button = (props: IButtonProps) => {
   return (
     <ConfigProvider>
-      <AntButton {...props}>{props.children}</AntButton>
+      <AntButton {...props} className={props.icon ? 'u-display-flex u-align-items-center ' : ''}>
```
jared-dickman, line 10:
```suggestion
      <AntButton {...props} className={props.icon ? 'u-display-flex u-align-items-center' : ''}>
```

PR #244, opened 2024-05-24
fix: button alignment

gabyzif:
When tested the previous implementation on MP, I found out that when adding the new class, I removed the classes from the props that ants injects, that was causing the button not to be full width.

I also realized that it would be a better approach to add variants to the button so in case we need more classes in the future, it would be just a matter of adding a class into the map :) 

Screenshot 2024-05-24 at 1 22 52 PM


src/components/general/Button/Button.stories.tsx
```diff
@@ -116,9 +116,19 @@ export const WithIconSM: Story = {
   args: {
     type: 'default',
     icon: <Icon name="mpLogo" size="sm" />,
+    variant: 'content-align-center',
   },
 }
 
+export const CircleWithoutText: Story = {
```
gabyzif, line 123: This should be improved:

RoundIconButton?

tibuurcio: makes sense to me

src/components/general/Button/Button.tsx
```diff
@@ -2,12 +2,16 @@ import { Button as AntButton } from 'antd'
...
 import { ConfigProvider } from 'src/components/other/ConfigProvider/ConfigProvider'
 
-export interface IButtonProps extends AntButtonProps {}
-
+export interface IButtonProps extends AntButtonProps {
+  variant?: 'content-align-center'
+}
 export const Button = (props: IButtonProps) => {
+  const classMap = {
+    'content-align-center': 'u-display-flex u-align-items-center',
```
tibuurcio, line 10: I like the idea of creating variants, always. Like Gabe taught us, we should not allow a bunch of props, variants are the way to go.

In this case though, when are we going to not use this variant? I think you said it yesterday, like: why would we ever want a not aligned button with icon?

PR #245, opened 2024-05-24
chore: add table export types

jared-dickman:
## Instructions

## Summary
- add table export types from ant


tibuurcio (approved):
👍🏼

PR #246, opened 2024-05-27
feat: Add links to the minimap

gabyzif:
This PR adds an SVG linker (like the one we have on MP for the overview map). Design gave us an SVGs with IDs for the minimap and with this component we are making the boxes clickable and with hover effects)

Note: I tried importing the SVG from a file like we do with the icons but its not working because the children IDs from the are not being passed. I'm going to investigate it in a TD task.

Note: this PR is under development.


src/components/navigation/GlobalNavigation/GlobalNavigationItems.d.ts
```diff
@@ -28,3 +28,17 @@ export interface IGlobalNavigationLink extends IBaseGlobalNavigationItem {
...
+  routes: ISvgLink[]
+}
+
+interface IMinimapOptionsHide {
+  show: false
+  overviewHref: undefined
+  routes: []
+}
+
+export type IMinimapOptions = IMinimapOptionsShow | IMinimapOptionsHide
```
tibuurcio, line 44: Interesting, I like it! ✨ 

suggestion to rename `Show` and `Hide` in the interface to `Shown` and `Hidden` just to make it more clear on the intention of the interface.

tibuurcio (commented):
It's looking good! Left 1 small suggestion :)

src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
```diff
@@ -486,12 +483,22 @@ export const MP: Story = {
     tools: mpTools,
     management: mpManagement,
     orgs: mpOrgs,
-    minimapOptions: { overviewHref: '/', show: true },
+    minimapOptions: {
+      overviewHref: '/',
+      handleLinkClick: string => alert(string),
```
tibuurcio, line 488: **suggestion** let's not use `string` as a parameter name not to confuse with the `string` type. `text` should be better here.

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -37,10 +41,7 @@ export interface IGlobalNavigationProps {
     onClick: () => void
     withoutContainer?: boolean
   }
-  minimapOptions: {
-    overviewHref?: string
-    show?: boolean
-  }
+  minimapOptions?: IMinimapOptions
```
tibuurcio, line 44: praise ✨ 
makes much more sense

src/components/navigation/GlobalNavigation/GlobalNavigationItems.d.ts
```diff
@@ -31,3 +32,5 @@ export interface IGlobalNavigationLink extends IBaseGlobalNavigationItem {
 }
 
 export type IGlobalNavigationItem = IGlobalNavigationMenu | IGlobalNavigationLink
+
+export type IMinimapOptions = { overviewHref: string; routes: ISvgLink[]; handleLinkClick: (route: string) => void }
```
tibuurcio, line 36: **question** `handleLinkClick`: by only looking at it, which link are we referring to? Is it all the routes? What about `handleRoutesClick`?

Also, if those routes are predefined we should be able to type this `route: string` better 🙏🏼

src/components/navigation/MiniMap/SvgLinker.tsx
```diff
@@ -0,0 +1,90 @@
+import React, { Children, ReactElement, useCallback } from 'react'
+import './miniMap.css'
+
+export interface ISvgLink {
+  elementId: string
+  route: string
+  isAuthorized: boolean
```
tibuurcio, line 7: @gabrielmarquesf Do you thin we will ever need a 3rd status here? Like `isAuthorized` is currently `boolean`, but could we ever need something like `permission: 'authorized' | 'not-authorized' | 'view-only' | 'blocked'` (only examples).

Trying to think if it makes sense to get rid of the boolean and have type instead :)

src/components/navigation/MiniMap/SvgLinker.tsx
```diff
@@ -0,0 +1,90 @@
...
+}
+
+interface ISvgLinkerProps {
+  buttons: ISvgLink[]
+  children: React.ReactNode
+  onLinkClick: (route: string) => void
+}
+
+export const SvgLinker: React.FC<ISvgLinkerProps> = ({ buttons, children, onLinkClick }) => {
+  const handleContainerClick = useCallback(
```
gabrielmarquesf, line 18: question: there is a reason to use `useCallback` here? I can't see this variable part of any memoization, but Idk if there is another usecase for it

src/components/navigation/MiniMap/SvgLinker.tsx
```diff
@@ -0,0 +1,90 @@
...
+                className={`svg-linker-root__button svg-linker-root__button--${button.variant}${
+                  button.isAuthorized ? '' : ' svg-linker-root__button--disabled'
+                }`}
+                onClick={e => {
+                  e.preventDefault()
+                  if (button.isAuthorized) {
+                    onLinkClick(button.route)
+                  } else {
+                    alert('You are not authorized to access this page.')
+                  }
```
gabrielmarquesf, line 63: you shouldn't need this since you have the `handleContainerClick` attached to the container, right?

src/components/navigation/MiniMap/SvgLinker.tsx
```diff
@@ -0,0 +1,90 @@
...
+          return React.cloneElement(element, { children: wrappedChildren })
+        }
+        return element
+      }
+    }
+
+    return Children.map(parent, child => {
+      if (!React.isValidElement(child)) return child
+      return wrapElement(child as ReactElement)
+    })
```
gabrielmarquesf, line 84: When I wrote this wrapper, I started with this same approach, but I moved to the version we currently have [in Nancy](https://git.corp.mparticle.com/mParticle/mPServer/blob/7be859413125e4025b481f5467eedd4b9d293556/src/UI/MP.UI.Web.Nancy/Assets/private/aurelia/src/viewmodels/overview-map/components/root-content-elements/OverviewMapRootContent.tsx#L32) because I believe it is much cleaner, I believe it's cleaner and it avoids having that `if (element.type === 'svg') {` block above. I would like to hear from you why you prefer this way

src/components/navigation/MiniMap/SvgLinker.tsx
```diff
@@ -0,0 +1,90 @@
...
+              </a>
+            )
+          }
+        }
+        if (element.props.children) {
+          const wrappedChildren = Children.map(element.props.children, child => {
+            if (!React.isValidElement(child)) return child
+            return wrapElement(child)
+          })
+          return React.cloneElement(element, { children: wrappedChildren })
```
gabrielmarquesf, line 75: this is identical to the code above inside the `if (element.type === 'svg') {`, makes sense to encapsulate it?

src/components/navigation/MiniMap/miniMap.svg
gabrielmarquesf, line 1: note, if you keep the inline version, remove this file

src/components/navigation/MiniMap/MiniMap.tsx
```diff
@@ -3,22 +3,455 @@ import './miniMap.css'
 import React from 'react'
 import { Button, ConfigProvider } from 'src/components'
 import Logo from 'src/assets/svg/mp-logo-wordmark.svg?react'
-import MiniMap from './miniMap.svg?react'
+import MiniMapSvg from './miniMap.svg?react'
 import { Flex } from 'src/components/layout/Flex/Flex'
+import { ISvgLink, SvgLinker } from 'src/components/navigation/MiniMap/SvgLinker'
+
+const minimap = (
```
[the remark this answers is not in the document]
gabyzif: So for some reason VIte doesn't handle correctly the props of the svg if import it as a file, that's why I put it in a variable

this is how the object looks when importing it as a file

Screenshot 2024-05-24 at 2 29 42 PM


vs this is how the object looks when using it as a variable.

Screenshot 2024-05-24 at 3 08 50 PM

for now, i'll add it in a different file but as a variable and crete a TD to check this issue (I did enter into a rabbit whole)

src/components/navigation/GlobalNavigation/GlobalNavigation.tsx
```diff
@@ -15,10 +15,14 @@ import { NavigationSearch } from 'src/components/navigation/GlobalNavigation/Nav
...
 import { WorkspaceSelector } from 'src/components/navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelector'
-import { type IGlobalNavigationItem } from 'src/components/navigation/GlobalNavigation/GlobalNavigationItems'
+import {
+  type IGlobalNavigationItem,
+  IMinimapOptions,
+} from 'src/components/navigation/GlobalNavigation/GlobalNavigationItems'
 import { NavigationItem } from 'src/components/navigation/GlobalNavigation/NavigationItem'
 import { useNewExperienceReminder } from 'src/hooks/NewExperienceReminder/useNewExperienceReminder'
 import { HomeButton } from 'src/components/navigation/GlobalNavigation/HomeButton'
+import { ISvgLink } from 'src/components/navigation/MiniMap/SvgLinker'
```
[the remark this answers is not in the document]
jared-dickman: run `optimize imports` in the ide

src/components/navigation/MiniMap/MiniMap.tsx
```diff
@@ -3,22 +3,455 @@ import './miniMap.css'
...
+  </svg>
+)
 export interface IMinimapProps {
   overviewHref: string
+  routes: { elementId: string; route: string; isAuthorized: boolean }[]
+  handleLinkClick: (route: string) => void
 }
 
-const Minimap: React.FC<IMinimapProps> = props => {
+const Minimap: React.FC<IMinimapProps> = ({ overviewHref, routes, handleLinkClick }) => {
```
jared-dickman, line 427:
```suggestion
const Minimap: React.FC<IMinimapProps> = (props: IMinimapProps) => {
```

src/components/navigation/MiniMap/miniMap.css
```diff
@@ -1,3 +1,62 @@
 .minimap_container{
```
jared-dickman, line 1: whats with this class name?!

src/components/navigation/MiniMap/SvgLinker.tsx
tibuurcio: hahaha I get it. Can be pre-engineering for sure. Problem is, if we ever decide to change we need to introduce a breaking change :/

So if there's only the slight possibility of having multiple status I would still go with an union string type.

src/components/navigation/GlobalNavigation/HomeButton.tsx
```diff
@@ -9,27 +11,30 @@ interface MpHomeButtonProps {
...
+
 const MpHomeButton: React.FC<MpHomeButtonProps> = ({ onClick }) => (
   <Center className="globalNavigation__mpHome" onClick={onClick}>
     <Icon name="mpLogo" size="lg" color="white" />
   </Center>
 )
 
-const MinimapWithPopover: React.FC<MinimapWithPopoverProps> = ({ overviewHref, onClick }) => (
-  <Popover content={() => <MiniMap overviewHref={overviewHref} />} placement="rightBottom" arrow={false}>
+const MinimapWithPopover: React.FC<MinimapWithPopoverProps> = ({ overviewHref, onClick, routes, handleLinkClick }) => (
```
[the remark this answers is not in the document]
tibuurcio: kudos for bringing our styleguide up. Although it should be an automated, we gotta stick to manually doing it while we're not there yet.

src/components/navigation/GlobalNavigation/GlobalNavigationItems.d.ts
gabrielmarquesf: > That's for prettier do decide. 😆 We have quite a large lineWidth configuration at 120, default was 80 but we tried to match better with current codebases.
(maybe the prettier rule would be `singleAttributePerLine` and not `lineWidth`, but I know how it's annoying with react comps with a few attrs)

Not sure if I agree this is up to prettier. It's not a stylish detail, I see some good technical reasons to break it

src/components/navigation/MiniMap/SvgLinker.tsx
```diff
@@ -0,0 +1,90 @@
...
+          onLinkClick(route)
+        } else {
+          alert('You are not authorized to access this page.')
+        }
+      }
+    },
+    [buttons, onLinkClick],
+  )
+
+  const wrapButtonsIntoLinks = (parent: React.ReactNode): React.ReactNode => {
```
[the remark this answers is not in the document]
gabyzif: yup. the idea at some point is that we use the linker in nancy too :) for the overview map

src/components/navigation/GlobalNavigation/GlobalNavigationItems.d.ts
tibuurcio: So @gabyzif added a new property to the `IMinimapOptions` and that seems to have solved the problem cause now Prettier is breaking the lines. But let me talk for a moment about formatting here.

> I know how it's annoying with react comps with a few attrs)

I actually don't find it annoying. For me, more than 2 properties is already enough for breaking, but that's personal opinion. Prettier doesn't allow this to be configurable, so that's what I meant with it's up for Prettier to decide. It means if we are using Prettier, we don't have a choice here 😅 (Other than disabling it and stop using it or going with another formatting tool)

If we had something more close to the default `lineWidth`, this would be solved, but we decided for `120` when we created the rules for the Aquarium so we sometimes end up with those long lines, as we would end up with sometimes too much new lines if we had the default at `80`.

It's harsh, and opinionated, but that's a good thing IMO cause we avoid this types of discussions on formatting.

> I see some good technical reasons to break it

Can you elaborate? Maybe you're thinking of readability and maintainability improvements? But that doesn't change anything technical about the code.

tibuurcio: Poor @gabyzif we love to use your PR's for discussion 😆

Just ignore the above in regards to the PR review please 🙏🏼

gabrielmarquesf: yes, you are right, by "technical reasons" I meant readability and maintainability. Besides the well known benefits, I raise how this makes historical changes tracking/finding easier, and it's very important imo.

And yes, makes sense to me everything you mentioned! ;)

src/components/navigation/GlobalNavigation/GlobalNavigationItems.d.ts
```diff
@@ -31,3 +31,17 @@ export interface IGlobalNavigationLink extends IBaseGlobalNavigationItem {
...
 
 export type IGlobalNavigationItem = IGlobalNavigationMenu | IGlobalNavigationLink
+
+export type MiniMapLinks = 'oversight' | 'dataPlatform' | 'customer360' | 'predictions' | 'analytics' | 'segmentation'
+export type MiniMapLink = {
+  linkId: MiniMapLinks
+  href: string
+}
+
+export interface IMinimapOptions {
```
gabrielmarquesf, line 41:
```suggestion
export interface IMiniMapOptions {
```

src/components/navigation/MiniMap/SvgLinker.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { Children, ReactElement } from 'react'
+import './miniMap.css'
+
+export interface ISvgLink {
+  elementId: string
+  href: string
+  variant?: 'regular' | 'black' | 'drop-shadow'
+  isUnAuthorized?: boolean
```
gabrielmarquesf, line 8:
```suggestion
  isUnauthorized?: boolean
```

src/components/navigation/MiniMap/miniMap.css
```diff
@@ -1,3 +1,61 @@
-.minimap_container{
+.minimap_container {
     padding: var(--padding-sm);
-}
\ No newline at end of file
+}
```
gabrielmarquesf, line 3: didn't work to replace it by a util class?

src/components/navigation/MiniMap/miniMap.css
```diff
@@ -1,3 +1,62 @@
...
+            fill: var(--color-text-disabled);
+        }
+
+        &:hover rect:first-child {
+            fill: transparent;
+        }
+    }
+
+    &.svg-linker-root__button--drop-shadow {
+        & > g > g > rect {
```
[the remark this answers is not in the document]
gabrielmarquesf: We need to sync it with UX, since they are exporting those SVGs for us. I will handle it!, and I think we can go ahead with this PR meanwhile.

PR #251, opened 2024-05-29
Temp Table - UNI-637

ajenkins-mparticle:
## Instructions

## Summary

## Testing Plan

### DIFF
--- a/src/components/data-display/TempTable/TempTable.tsx
+++ b/src/components/data-display/TempTable/TempTable.tsx
@@ -29,7 +29,7 @@ export type { ColumnType, ExpandableConfig, ColumnsType, TableProps }
 export const Table = <RecordType extends AnyObject>(props: ITableProps<RecordType>) => {
   return (
     <ConfigProvider>
-      <AntTable<RecordType>
+      <AntTable
         {...props}
         columns={props.columns.map(column => ({
           ...column,
--- a/src/components/data-display/TempTable/TempTable.tsx
+++ b/src/components/data-display/TempTable/TempTable.tsx
@@ -1,0 +1,62 @@
+import { Table as AntTable } from 'antd'
+import { type TableProps as AntTableProps } from 'antd'
+import { ConfigProvider } from 'src/components'
+import { type ExpandableConfig } from 'antd/es/table/interface'
+import type { AnyObject } from 'antd/es/_util/type'
+import type { ColumnGroupType, ColumnType, ColumnsType, TableProps } from 'antd/es/table'
+
+// example
+export type SupportedTableType = 'text' | 'date'
+
+export type IColumnsGroupType<RecordType> = ColumnGroupType<RecordType> & {
+  type?: SupportedTableType
+}
+
+export type IColumnType<RecordType> = ColumnType<RecordType> & {
+  type?: SupportedTableType
+}
+
+export type IColumnsType<RecordType> = Array<IColumnType<RecordType> | IColumnsGroupType<RecordType>>
+
+export type ITableColumn<RecordType> = IColumnsType<RecordType>
+
+export interface ITableProps<RecordType> extends Omit<AntTableProps<RecordType>, 'columns'> {
+  columns: ITableColumn<RecordType>
+}
+
+export type { ColumnType, ExpandableConfig, ColumnsType, TableProps }
+
+export const Table = <RecordType extends AnyObject>(props: ITableProps<RecordType>) => {
+  return (
+    <ConfigProvider>
+      <AntTable<RecordType>
+        {...props}
+        columns={props.columns.map(column => ({
+          ...column,
+          render: (...args): React.ReactNode => {
+            if (!column.type) {
+              console.warn("Should've specified a type, attempting raw render")
+              if (!column.render) throw new Error('No type and no render specified')
+              return <>{column.render(...args)}</>
+            }
+            if (column.type === 'date') {
+              return <>Date formatted value {args[0]}</>
+            }
+
+            return <>String formatted val {args[0]}</>
+          },
+        }))}
+      />
+    </ConfigProvider>
+  )
+}
+
+Table.Column = AntTable.Column
+Table.ColumnGroup = AntTable.ColumnGroup
+Table.Summary = AntTable.Summary
+
+Table.EXPAND_COLUMN = AntTable.EXPAND_COLUMN
+Table.SELECTION_NONE = AntTable.SELECTION_NONE
+Table.SELECTION_ALL = AntTable.SELECTION_ALL
+Table.SELECTION_INVERT = AntTable.SELECTION_INVERT
+Table.SELECTION_COLUMN = AntTable.SELECTION_COLUMN
--- a/src/components/data-display/TempTable/TempTable.stories.tsx
+++ b/src/components/data-display/TempTable/TempTable.stories.tsx
@@ -1,0 +1,23 @@
+import { type Meta, type StoryObj } from '@storybook/react'
+import { Table } from 'src/components/data-display/TempTable/TempTable'
+
+const meta: Meta<typeof Table> = {
+  title: 'Aquarium/Data Display/TempTable',
+  component: Table,
+
+  args: {},
+}
+export default meta
+
+type Story = StoryObj<typeof Table>
+
+export const Primary: Story = {
+  args: {
+    columns: [
+      {
+        title: 'Hi',
+        type: 'date',
+      },
+    ],
+  },
+}


src/components/data-display/TempTable/TempTable.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { type TableProps as AntTableProps } from 'antd'
+import { ConfigProvider } from 'src/components'
+import { type ExpandableConfig } from 'antd/es/table/interface'
+import type { AnyObject } from 'antd/es/_util/type'
+import type { ColumnGroupType, ColumnType, ColumnsType, TableProps } from 'antd/es/table'
+
+// example
+export type SupportedTableType = 'text' | 'date'
+
+export type IColumnsGroupType<RecordType> = ColumnGroupType<RecordType> & {
```
Seebiscuit, line 11: **suggestion** Not sure we need this now. If we don't have a use-case for this, let's get rid of it.

src/components/data-display/TempTable/TempTable.tsx
```diff
@@ -0,0 +1,62 @@
...
+  type?: SupportedTableType
+}
+
+export type IColumnType<RecordType> = ColumnType<RecordType> & {
```
Seebiscuit, line 15: I'm hoping not to extend `ColumnType` and instead build our own from scratch based on the immediate use-cases and props that we consider reasonably useful. That's why studying the Observability mocks and the Column API would be useful. I'm ok with you making educated guesses here. I'm especially sensitive to props that provide too much flexibility.

Seebiscuit, line 15: **thought** While considering the API for `Tag` I realized that the majority of the config will be in the `DataType`. For example, this is what the API will look like:

First, some simple foundational interfaces:
```typescript
interface BaseData {
    key: number | string,
}

enum ColumnType {
    Text = 'text',
    Tag = 'tag',
}

interface Column<DataType> {
    title: string,
    type: ColumnType,
    dataIndex: string,
    key: string,
    render?: (data: DataType) => React.ReactNode,
}

interface TableProps<DataType> {
    columns: Column<DataType>[],
    dataSource: DataType[],
}
```

And now how we would build the interface to implement Tag:
```typescript
enum TagType {
    primary = 'primary',
    success = 'success',
    error = 'error',
    warning = 'warning',
}

interface TagData {
    type: TagType,
    text: string,
}

interface ExampleData extends BaseData {
    name: string,
    age: number,
    tag: TagData,
}

const columns: Column<ExampleData>[] = [
    {
        title: 'Name',
        dataIndex: 'name',
        key: 'name',
        type: ColumnType.Text
    },
    {
        title: 'Age',
        dataIndex: 'age',
        key: 'age',
        type: ColumnType.Text
    },
    {
        title: 'Address',
        dataIndex: 'tag',
        key: 'tag',
        type: ColumnType.Tag
    },
];

const dataSources: ExampleData[] = [
    {
        key: 1,
        name: 'John Brown',
        age: 32,
        tag: {
            type: TagType.primary,
            text: 'New York No. 1 Lake Park',
        },
    },
    {
        key: 2,
        name: 'Jim Green',
        age: 42,
        tag: {
            type: TagType.success,
            text: 'London No. 1 Lake Park',
        },
    }
];
```

(Please, don't mind me playing fast and loose with naming conventions.)

src/components/data-display/TempTable/TempTable.tsx
```diff
@@ -0,0 +1,62 @@
...
+  columns: ITableColumn<RecordType>
+}
+
+export type { ColumnType, ExpandableConfig, ColumnsType, TableProps }
+
+export const Table = <RecordType extends AnyObject>(props: ITableProps<RecordType>) => {
+  return (
+    <ConfigProvider>
+      <AntTable
+        {...props}
```
Seebiscuit, line 33: **suggestion** A bit concerned by passing `{...props}` in general. Would prefer to be specific about how we're using the component.

src/components/data-display/TempTable/TempTable.tsx
ajenkins-mparticle: Right now we're creating column types that we see in the observability figmas only. As new UI patterns are created/seen, how will we know which ones are render function worthy vs column config worthy?

Let's say we have a column that renders 2 types of tags, or maybe even n tags - like "this record is tagged with these tags"

Would we add a 

```
getTagTypes?: (tags: string[], record: DataType, index: number) => TagType[],
```

or would we add:

```
getTagTypeForTagAtIndex?: (tag: string, record: DataType, rowIndex: number, indexOfTag: number) => TagTag`
```

This exact use case isn't super important, just trying to think out loud about how this scales

src/components/data-display/TempTable/TempTable.tsx
ajenkins-mparticle: antd gives us so much config that's desired by design out of the box (expandable rows have [documented examples](https://ant.design/components/table#components-table-demo-expand), just for one example).

No problem starting with just `columns` and `dataSource`. Would like to make the callout that it's likely we're going to be adding back in things that design will use in their antd-based figma design system (requiring aquarium releases for consumers to actually use them) because we decided not to include them in the first place.

ajenkins-mparticle: Want to circle back to this one, @Seebiscuit - what are the desired props we should accept for the first iteration?

src/components/data-display/TempTable/TempTable.tsx
ajenkins-mparticle: @Seebiscuit there's a couple of code examples of yours above that differ, looks like I should implement the second one, is that correct?

Seebiscuit: On the first point about returning `TagType[]` (instead of `TagType`), I think that's a valid point and we should go with `TagType[]`. 

But this pattern may or may not translate to other standard candidates (like Status, etc). We should be sensitive to context (ie, usage).

Seebiscuit: On the point of which variation to use, I'm starting to like the second variant, where the Column holds all the config. What do you think?

Seebiscuit: > Right now we're creating column types that we see in the observability figmas only. As new UI patterns are created/seen, how will we know which ones are render function worthy vs column config worthy?

I misunderstood your question. It's a good one. 

There's the concept of a standards candidate. That means it's a piece of UI that Design has deemed to be universal and reusable enough to be included in the Aquarium. Once it makes it into the Aq it becomes a standards component, or just a component. 

I've been borrowing that concept for the Table. Speaking with Sankalp and using his exploratory docs on the Table Component, he's marked out the pieces that should be elevated to "standard components". These are the pieces we should use the Column config for. 

The render fn, then would be used for non-standard work. And, hence, the idea to emit the warning message to let us know it's usage and to understand if there's anything there that can be elevated to a standards candidate.

Seebiscuit: So, I been sleeping on this comment. I think we should discuss it more.

On the one hand, creating a curated props list is nice. On the other hand, I'm loath to recreate it if we're going to end up including 90%+ of what's in there. (Same goes for the Table prop.)

What do you think? I'd appreciate a discussion here.

src/components/data-display/TempTable/TempTable.tsx
Seebiscuit: So, this is separate from our interface discussion about what props should go here. This is more about being deliberate with what props are passed to the table. It's more about exposing visibility about what goes into the component and preventing hacks that rely on unpublished interfaces. This is a principle the UI decide to adopt as a team.

ajenkins-mparticle: No problem, don't spread props, that's fine. Still open ended as to what our initial props we're desiring to pass into the table - do we start with just `columns` and `dataSource`? Are there any others we want from [here](https://ant.design/components/table#table)?

src/components/data-display/TempTable/TempTable.tsx
Seebiscuit: >On the first point about returning TagType[] (instead of TagType), I think that's a valid point and we should go with TagType[].

I just realized that this will be complicated. The expectation is to get a single tag and decorate it with a configured `<Tag />`.  If we want to return a `TagType[]` we are going to have to have docs explaining what the `split` char will be. But, you're probably right that it makes sense to support multiple tag. No need to solution now, that's what the Tag work item will be for.

jared-dickman:
cc @brewdente

src/components/data-display/TempTable/TempTable.tsx
ajenkins-mparticle: Ok, excellent, so do we align on standards candidates at this stage? If so, what would the standards candidates be?

I'm guessing:

- `tag`
- `date`
- `link` - looks like there' s a clickable link component in the observability figma
- `text` (default?)
- there's a cell with 2 lines - what would the column configuration would that look like?


[Two line cell:](https://www.figma.com/design/hDSxMHoLlMlLeYbgvT3A6d/Observability-Exploration?node-id=2825-271478&m=dev)

image

[Link Cell](https://www.figma.com/design/hDSxMHoLlMlLeYbgvT3A6d/Observability-Exploration?node-id=2825-271477&m=dev)

image

jared-dickman:
oh! it looks like you opened a pr against a forked version of the repo, rather than a cloned

i think we're advising for internal devs to clone, because when forked it means that i cannot check out this branch locally to run/debug/review

jared-dickman:
@ajenkins-mparticle ah totally makes sense!
you can bump your ticket your tag sam/yuan to expedite it if you dont hear anything in the coming time

image

src/components/data-display/TempTable/TempTable.tsx
Seebiscuit: I highlighted the standard candidates in light green on the [Table Project Plan](https://docs.google.com/spreadsheets/d/1AaFcG7f1ZbLwE4brTB1ja9N3evu2Ef83W6b04VFep3g/edit?usp=sharing) (this includes the Compose column type, in case you're thinking about the interface).

Good question on the Two Line cell (`TextDescription` in the Project Plan---not married to that name). My preference for that is to provide an interface like:

```typescript
interface Column<T> {
    //... all other props
    // type: ColumnType.TextDescription // -- documenting the type for the purpose of this discussion
    textDescriptionIndex: { textIndex: string, descriptionIndex: string }
}
```

I'm not loving this, something about it doesn't feel right. But it's a nod to the Antd way of using `dataIndex` to map a column to a data record property. 

In general, with these questions my answer will be, first, let's try to do it the way Antd would. As a veteran Antd user I'm hoping you'll help me refine the design of this component.

PR #253, opened 2024-05-31
feat: unnesting minimap styles

gabyzif:
## Instructions

## Summary

## Testing Plan

### DIFF
--- a/src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
+++ b/src/components/navigation/GlobalNavigation/GlobalNavigation.stories.tsx
@@ -486,7 +486,7 @@ export const MP: Story = {
     minimapOptions: {
       overviewHref: '/',
       onLinkClick: link => alert(link.href),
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## [1.16.1-minimap-styles.1](https://github.com/mParticle/aquarium/compare/v1.16.0...v1.16.1-minimap-styles.1) (2024-05-31)
+
+
+### Bug Fixes
+
+* in front of it ([97daf62](https://github.com/mParticle/aquarium/commit/97daf62cd5302862950d1bd60e999b20d5edddb9))
```
tibuurcio, line 6: Whatever we use as the commit message gets shown in the changelog and as we can see here it doesn't tell us much about what's going on 😅 

Can we somehow edit this message or even push a new commit explaining what is the fix?

tibuurcio (changes requested):

tibuurcio (approved):

PR #259, opened 2024-06-03
chore: add generic typing to table component

jared-dickman:
## Instructions

## Summary
- Adds generic template type for use in platforms 

## Testing Plan

- Closes https://go.mparticle.com/work/UI3DM-1245

### DIFF
--- a/src/components/data-display/Table/Table.tsx
+++ b/src/components/data-display/Table/Table.tsx
@@ -1,6 +1,6 @@
 import { Table as AntTable } from 'antd'
 import { type TableProps as AntTableProps } from 'antd'
-import { type AnyObject } from 'antd/lib/_util/type'
+import { type AnyObject } from 'antd/es/_util/type'
 import { ConfigProvider } from 'src/components'
 import { type ColumnType, type ExpandableConfig } from 'antd/es/table/interface'
 import { type ColumnsType, type TableProps } from 'antd/es/table'
--- a/src/components/data-display/Table/Table.tsx
+++ b/src/components/data-display/Table/Table.tsx
@@ -5,11 +5,11 @@ import { ConfigProvider } from 'src/components'
 import { type ColumnType, type ExpandableConfig } from 'antd/es/table/interface'
 import { type ColumnsType, type TableProps } from 'antd/es/table'
 
-export interface ITableProps<RecordType extends AnyObject = any> extends AntTableProps<RecordType> {}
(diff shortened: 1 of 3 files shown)


src/components/data-display/Table/Table.tsx
```diff
@@ -1,17 +1,18 @@
...
 import { type ColumnType, type ExpandableConfig } from 'antd/es/table/interface'
 import { type ColumnsType, type TableProps } from 'antd/es/table'
 
-export interface ITableProps extends AntTableProps {}
+export interface ITableProps<RecordType extends AnyObject = any> extends AntTableProps<RecordType> {}
 
 export type { ColumnType, ExpandableConfig, ColumnsType, TableProps }
 
-export const Table = (props: ITableProps) => {
+export const Table = <RecordType extends AnyObject = any>(props: ITableProps<RecordType>) => {
```
tibuurcio, line 12: Oh boy `AnyObject` that has `any` as a default. That's too many `any`'s 😄 

Jokes aside, can we do the [same as Antd](https://github.com/ant-design/ant-design/blob/c4147994907f313780a4721ca470bfde240350c6/components/table/Table.tsx#L18) and use `AnyObject = AnyObject`?

jared-dickman: sure thing, i got this from node_modules/antd/es/table/InternalTable.d.ts:10 but your link seems legit
image

ajenkins-mparticle (approved):
I was playing in Nancy and I found I needed this two Fridays ago. LGTM

src/components/data-display/Table/Table.tsx
```diff
@@ -1,17 +1,18 @@
 import { Table as AntTable } from 'antd'
 import { type TableProps as AntTableProps } from 'antd'
+import { type AnyObject } from 'antd/lib/_util/type'
```
ajenkins-mparticle, line 3: Does it matter if we mix imports from antd/es and antd/lib? I know this is only a type, I'm wondering how this transpiles, though. I was reading about [type imports](https://github.com/microsoft/TypeScript/issues/39861#issuecomment-2029976414) the other day and I think this transpiles to:

```
import { } from 'antd/lib/_util/type
```

This is minor, just asking if you knew how it all works, I don't really.

jared-dickman: i also dont really 😩 but we seem to mostly use /es
ive changed this and the one other /lib import to be consistent with the rest of the project 

image

tibuurcio (approved):

PR #260, opened 2024-06-03
Temp Table - SQDUI-1803

ajenkins-mparticle:
## Instructions

## Summary

## Testing Plan

- Closes https://go.mparticle.com/work/SQDUI-1803

### DIFF
--- a/src/components/data-display/TempTable/AntTable.stories.tsx
+++ b/src/components/data-display/TempTable/AntTable.stories.tsx
@@ -1,0 +1,91 @@
+import { faker } from '@faker-js/faker'
+import { type Meta, type StoryObj } from '@storybook/react'
+import { Table as TempTable, type TableColumnType } from 'antd'
+import { type ObservabilityRecord, createMockObservabilityRecord } from './observability.mock'
+import { TextCell, TextDescriptionCell } from './cells'
+import { Badge } from '../Badge/Badge'
+import { Typography } from 'src/components/general/Typography/Typography'
+import { Tag } from '../Tag/Tag'
+import { ConfigProvider } from 'src/components/other/ConfigProvider/ConfigProvider'
+
+const meta: Meta<typeof TempTable> = {
+  title: 'Aquarium/Data Display/TempAntTable',
+  component: TempTable,
+  decorators: [
+    Cmp => (
+      <ConfigProvider>
+        <Cmp />
+      </ConfigProvider>
+    ),
+  ],
+  args: {},
+}
+export default meta
+
+type Story = StoryObj<typeof TempTable<ObservabilityRecord>>
+
+const observabilityRecords = faker.helpers.multiple(createMockObservabilityRecord, { count: 10 })
+
+const columns: Array<TableColumnType<ObservabilityRecord>> = [
+  {
+    title: 'Request Timestamp',
+    key: 'timestamp',
+    dataIndex: 'timestamp',
+    render: (timestamp: string) => <TextCell text={new Date(timestamp).toISOString()} />,
+  },
+  {
+    title: 'Trace ID',
+    dataIndex: 'traceId',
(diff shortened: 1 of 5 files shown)


src/components/data-display/TempTable/AntTable.stories.tsx
```diff
@@ -0,0 +1,83 @@
+import { faker } from '@faker-js/faker'
```
ajenkins-mparticle, line 1: This is a demo story of what the developer usage would look like if we used straight antd table (decorated with the design tokens via the ConfigProvider)

src/components/data-display/TempTable/TempTable.stories.tsx
```diff
@@ -0,0 +1,94 @@
+import { faker } from '@faker-js/faker'
```
ajenkins-mparticle, line 1: This story is a demo of what the table would look like if we used a customized API that restricted the cell types

src/components/data-display/TempTable/TempTable.tsx
```diff
@@ -0,0 +1,121 @@
...
+  render: (value: any, data: DataType, index: number) => React.ReactNode
+}
+
+interface BaseColumnWithoutTypeAndRender<DataType> extends Omit<BaseColumn<DataType>, 'type' | 'render'> {}
+
+type ColumnGetProps<Props, DataType> = (dataType: DataType) => Props
+
+interface TagsColumn<DataType> extends BaseColumnWithoutTypeAndRender<DataType> {
+  type: 'tag'
+  getProps: ColumnGetProps<ITagCellProps, DataType>
```
ajenkins-mparticle, line 32: The idea is that we specify cell types - data, link, text, tag, etc and we provide a `getProps` function that maps the dataSource record to the props required by the specific component type.

I was pretty liberal with the props I allowed to be specified to the cells - see cells.tsx - but there's no problem to lock them down if we want. The main idea here is to bring the API of `type` and `getProps` out into the open for discussion

src/components/data-display/TempTable/observability.mock.ts
```diff
@@ -0,0 +1,29 @@
+import { faker } from '@faker-js/faker'
```
ajenkins-mparticle, line 1: Observability mocks for demonstration, not relevant to this PR

src/components/data-display/TempTable/cells.tsx
```diff
@@ -0,0 +1,56 @@
+import { type ILinkProps, Typography, type ITextProps } from 'src/components/general/Typography/Typography'
```
ajenkins-mparticle, line 1: These are a bunch of standard cell types. We can lock the props down further if we feel like we need to

src/components/data-display/TempTable/TempTable.tsx
```diff
@@ -0,0 +1,121 @@
...
+      <AntTable
+        pagination={pagination}
+        dataSource={dataSource}
+        columns={columns.map(column => ({
+          ...column,
+          render: (...args): React.ReactNode => {
+            const dataType = args[1]
+            if (!('type' in column)) {
+              console.warn('No type given, prefer to use a column type')
+              if (!('render' in column)) {
```
Seebiscuit, line 85: **question** In Antd you don't have to specify a render fn if you provide a `dataIndex` are we considering dropping the `dataIndex`?

src/components/data-display/TempTable/TempTable.tsx
ajenkins-mparticle: EDIT: Found a way around this, if the user doesn't supply a `type` they'll need to supply the `dataIndex` or the `render` function (and can supply both) - otherwise we'll throw an error.

https://github.com/mParticle/aquarium/pull/260/commits/8b4fe3393c14921eb60a95e49f47467f7453823c

PR #261, opened 2024-06-03
feat: allows UPS to receive cookie config options

tibuurcio:
## Summary
- We need to configure the `domain` option for the cookie to be able to be share user preferences between platforms. This PR allows the passing of a CookieConfig options object to the UPS.

## Testing Plan
- Release a feature version and use it under the same domain to test UPS cookie is shared between them

- Closes https://go.mparticle.com/work/UNI-646

### DIFF
--- a/src/services/user-preferences/user-preferences.ts
+++ b/src/services/user-preferences/user-preferences.ts
@@ -1,10 +1,10 @@
 /* eslint-disable @typescript-eslint/no-extraneous-class,no-unused-vars,@typescript-eslint/no-unused-vars */
-import * as Cookies from '../../utils/Cookies'
 import { type UserPreferences } from 'src/services/user-preferences/models/storage-models/user-preferences'
 import { type CompositeUserPreferences } from 'src/services/user-preferences/models/user-preferences/composite-user-preferences'
 import { type UserPreferenceScope } from 'src/services/user-preferences/models/storage-models/user-preference-scope'
 import { type UserPreferenceDefinitions } from 'src/services/user-preferences/models/definitions/user-preference-definitions'
 import { type CompositeUserPreferencesService } from 'src/services/user-preferences/composite-user-preferences-service'
+import * as Cookies from 'src/utils/Cookies'
 import { type CookieOptions } from 'src/utils/Cookies'
 
 export class UserPreferencesService<TUserPreferenceId extends PropertyKey> {
@@ -73,9 +73,7 @@ export class UserPreferencesService<TUserPreferenceId extends PropertyKey> {
   }
 
   private async setStoredPreferences(storedPreferences: UserPreferences<TUserPreferenceId>): Promise<void> {
-    Cookies.putObject(this.cookieOptions.key, storedPreferences, {
-      ...this.cookieOptions,
-    })
+    Cookies.putObject(this.cookieOptions.key, storedPreferences, this.cookieOptions)
 
     await Promise.resolve()
   }
--- a/src/services/user-preferences/user-preferences.ts
+++ b/src/services/user-preferences/user-preferences.ts
@@ -5,6 +5,7 @@ import { type CompositeUserPreferences } from 'src/services/user-preferences/mod
 import { type UserPreferenceScope } from 'src/services/user-preferences/models/storage-models/user-preference-scope'
 import { type UserPreferenceDefinitions } from 'src/services/user-preferences/models/definitions/user-preference-definitions'
 import { type CompositeUserPreferencesService } from 'src/services/user-preferences/composite-user-preferences-service'
+import { type CookieOptions } from 'src/utils/Cookies'
 
 export class UserPreferencesService<TUserPreferenceId extends PropertyKey> {
   public preferences!: CompositeUserPreferences<TUserPreferenceId>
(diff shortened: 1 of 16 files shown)


src/services/user-preferences/user-preferences.ts
```diff
@@ -5,16 +5,16 @@ import { type CompositeUserPreferences } from 'src/services/user-preferences/mod
...
 
 export class UserPreferencesService<TUserPreferenceId extends PropertyKey> {
   public preferences!: CompositeUserPreferences<TUserPreferenceId>
 
   constructor(
     private readonly definitions: UserPreferenceDefinitions<TUserPreferenceId>,
     private readonly compositeUserPreferencesService: CompositeUserPreferencesService<TUserPreferenceId>,
-    private readonly cookieKey: string,
     private readonly currentScope: UserPreferenceScope,
-    public dateFormatter: () => Date,
```
tibuurcio, line 17: No need for dateFormatter anymore cause that was used for the `expires` option which is now configurable from the client.

gabyzif (approved):

src/utils/Cookies.ts
```diff
@@ -12,23 +12,31 @@ export function getObject(key: string): string | null {
...
+  const defaultExpires = 'Thu, 01 Jan 1970 00:00:01 GMT'
   let expires = options.expires
-  if (value == null) expires = 'Thu, 01 Jan 1970 00:00:01 GMT'
+  if (value == null) expires = defaultExpires
   if (typeof expires === 'string') expires = new Date(expires)
   let str = `${_encode(key)}=${value != null ? _encode(value) : ''}`
   if (options.path) str += `; path=${options.path}`
   if (options.domain) str += `; domain=${options.domain}`
-  if (options.expires) str += `; expires=${expires.toUTCString()}`
+  if (options.expires) str += `; expires=${expires?.toUTCString() ?? defaultExpires}`
```
Seebiscuit, line 30: **praise** Much better.

src/utils/Cookies.ts
```diff
@@ -12,23 +12,31 @@ export function getObject(key: string): string | null {
...
+  domain?: string
+  expires?: string | Date
+  secure?: boolean
+}
+
+export function put(key: string, value: string | null, options: CookieOptions = {}): void {
+  const defaultExpires = 'Thu, 01 Jan 1970 00:00:01 GMT'
   let expires = options.expires
-  if (value == null) expires = 'Thu, 01 Jan 1970 00:00:01 GMT'
+  if (value == null) expires = defaultExpires
```
Seebiscuit, line 25: **question** Not sure I understand the business requirement here. Comments would be good. A private function encapsulating the business rule would be good, too.

src/utils/Cookies.ts
```diff
@@ -12,23 +12,31 @@ export function getObject(key: string): string | null {
   return value ? JSON.parse(value) : value
 }
 
-export function put(key: string, value: string | null, options: any /* TODO fix any */ = {}): void {
+export type CookieOptions = {
+  path?: string
+  domain?: string
+  expires?: string | Date
```
gabrielmarquesf, line 18: question: string is for iso format? maybe restricting to Date only would make it clear for the consumers.

gabrielmarquesf (approved):

src/utils/Cookies.ts
Seebiscuit, line 18: **thought** Would be good to have a flag, like `permanent` that sets the cookie for, say, 100 years. That way the client doesn't always have to calculate the "permanent" date.

src/utils/Cookies.ts
tibuurcio: Good question. I don't think I understand it either 😅 This is [copied from the aurelia-plugins-cookies](https://github.com/aurelia-plugins/aurelia-plugins-cookies/blob/5566e5bb55d2f56d9cd930ba348cac337971aeb7/src/aurelia-plugins-cookies.js#L20) codebase.

By just looking at the code, I think there's 2 options: either the cookie is born invalid already or it never expires. Running some tests to see.

src/utils/Cookies.ts
tibuurcio: Agree. Yeah string is for iso format cause [the code where this came from originally](https://github.com/aurelia-plugins/aurelia-plugins-cookies/blob/5566e5bb55d2f56d9cd930ba348cac337971aeb7/src/aurelia-plugins-cookies.js#L20) supported both.

src/utils/Cookies.ts
tibuurcio: So I figured it out by testing the implementation directly in the browser.

Turns out this defaultExpires is only used for the `remove` method since it doesn't actually "remove" the cookie but rather set it's value to `null` and call this `put` method. This makes it so that the expiration date goes to the default one and invalidates the cookie.

src/utils/Cookies.ts
tibuurcio: I actually made it only support string given that's what the cookie expires option needs to be set to.

tibuurcio: I like it, although I don't think we should be doing Date manipulation inside the cookie library so I'm proposing we set it to a fixed value in the future.

src/utils/Cookies.ts
```diff
@@ -12,23 +14,58 @@ export function getObject(key: string): string | null {
...
+export type CookieOptions = RequireOneOrNone<
+  {
+    path?: string
+    domain?: string
+    secure?: boolean
+    expiresISOString: string
+    permanent: boolean
+  },
+  'expiresISOString' | 'permanent'
+>
```
tibuurcio, line 26: If we use permanent, expires is forced as undefined
Screenshot 2024-06-07 122402

If we use expires, permanent is forced as undefined
Screenshot 2024-06-07 122434

Not allowed to use both
Screenshot 2024-06-07 122525

Even in JS land, precedence would be given to `permanent`

tibuurcio:
Tested this implementation both in Nancy and Indicative codebases using a feature version.

src/utils/Cookies.ts
gabrielmarquesf: 🔥 nice

gabrielmarquesf (approved):

gabyzif (approved):

PR #262, opened 2024-06-04
feat: Add size prop to `Typography.Text` component

ajenkins-mparticle:
## Instructions

## Summary

- [EAMES](https://www.figma.com/design/LffDbOUjeYqDMZ3djs9Cga/mParticle-Foundation-v1.0.1?node-id=3745-8164&m=dev) specifies a size for a `Text` component. 

Screenshot 2024-06-04 at 8 15 36 AM

## Testing Plan

- Tested in storybook, using a control. Default behaviour (`base`) is preserved so it's backwards compat.

### DIFF
--- a/src/components/general/Typography/Typography.tsx
+++ b/src/components/general/Typography/Typography.tsx
@@ -25,6 +25,8 @@ export interface ITextProps extends AntTextProps {
   size?: TypographySize
 }
 
+// TODO: Replace hardcoded values in getFontSize and getLineHeight with tokens when design is ready
+// These values are currently coming from https://www.figma.com/design/LffDbOUjeYqDMZ3djs9Cga/mParticle-Foundation-v1.0.1?node-id=3745-8164&m=dev
 const getFontSize = (size: TypographySize): number => {
   if (size === 'base') return 14
   if (size === 'sm') return 12
--- a/src/components/general/Typography/Typography.tsx
+++ b/src/components/general/Typography/Typography.tsx
@@ -1,10 +1,18 @@
-import { Typography as AntTypography, type TypographyProps as AntTypographyProps } from 'antd'
+import {
+  Typography as AntTypography,
+  type TypographyProps as AntTypographyProps,
+  theme,
+  ConfigProvider as AntConfigProvider,
+} from 'antd'
 import { ConfigProvider } from 'src/components'
 import { type ReactNode } from 'react'
 import { type TextProps as AntTextProps } from 'antd/es/typography/Text'
 import { type TitleProps as AntTitleProps } from 'antd/es/typography/Title'
 import { type LinkProps as AntLinkProps } from 'antd/es/typography/Link'
 import { type ParagraphProps as AntParagraphProps } from 'antd/es/typography/Paragraph'
+import type { FontMapToken } from 'antd/es/theme/interface'
+
+const { useToken } = theme
 
 export interface ITypographyProps extends AntTypographyProps {
   children: ReactNode
@@ -21,25 +29,33 @@ export interface ITextProps extends AntTextProps {
   size?: TypographySize
 }
 
-const getFontSizeVariable = (size: TypographySize): string => `--font-size${size === 'base' ? '' : `-${size}`}`
+type FontSizeToken = keyof Pick<FontMapToken, 'fontSize' | 'fontSizeSM' | 'fontSizeLG' | 'fontSizeXL'>
+type LineHeightToken = keyof Pick<FontMapToken, 'lineHeight' | 'lineHeightSM' | 'lineHeightLG'>
 
-const getLineHeightVariable = (size: TypographySize): string => {
-  if (size === 'sm') return `--line-height-sm`
-  if (size === 'lg' || size === 'xl') return '--line-height-lg'
-  return `--line-height`
+const getFontSizeToken = (size: TypographySize): FontSizeToken => {
+  if (size === 'base') return 'fontSize'
+  if (size === 'sm') return 'fontSizeSM'
+  if (size === 'lg') return 'fontSizeLG'
+  return 'fontSizeXL'
+}
+
+const getLineHeightToken = (size: TypographySize): LineHeightToken => {
+  if (size === 'base') return 'lineHeight'
+  if (size === 'sm') return 'lineHeightSM'
+  return 'lineHeightLG'
 }
 
 const Text = ({ size = 'base', ...props }: ITextProps) => {
+  const { token } = useToken()
+
+  const fontSize = token[getFontSizeToken(size)]
+  const lineHeight = token[getLineHeightToken(size)]
+
   return (
     <ConfigProvider>
-      <AntTypography.Text
-        style={{
-          fontSize: `var(${getFontSizeVariable(size)})`,
-          lineHeight: `var(${getLineHeightVariable(size)})`,
-        }}
-        {...props}>
-        {props.children}
-      </AntTypography.Text>
+      <AntConfigProvider theme={{ components: { Typography: { fontSize, lineHeight } } }}>
+        <AntTypography.Text {...props}>{props.children}</AntTypography.Text>
+      </AntConfigProvider>
     </ConfigProvider>
   )
 }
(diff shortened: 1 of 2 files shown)


src/components/general/Typography/Typography.tsx
```diff
@@ -16,14 +16,34 @@ export const Typography = (props: ITypographyProps) => (
...
+    <ConfigProvider>
+      <AntTypography.Text
+        style={{
+          fontSize: `var(${getFontSizeVariable(size)})`,
+          lineHeight: `var(${getLineHeightVariable(size)})`,
+        }}
+        {...props}>
+        {props.children}
+      </AntTypography.Text>
+    </ConfigProvider>
```
ajenkins-mparticle, line 43: An alternative to inline styling would've been nesting a config provider

```suggestion
    <ConfigProvider>
      <AntTypography.Text
        style={{
          fontSize: `var(${getFontSizeVariable(size)})`,
          lineHeight: `var(${getLineHeightVariable(size)})`,
        }}
        {...props}>
        {props.children}
      </AntTypography.Text>
    </ConfigProvider>
```

src/components/general/Typography/Typography.tsx
```diff
@@ -16,14 +24,47 @@ export const Typography = (props: ITypographyProps) => (
...
+}
+
+const getLineHeightToken = (size: TypographySize): keyof FontMapToken => {
+  if (size === 'base') return 'lineHeight'
+  if (size === 'sm') return 'lineHeightSM'
+  return 'lineHeightLG'
+}
+
+const Text = ({ size = 'base', ...props }: ITextProps) => {
+  const { token } = useToken()
```
ajenkins-mparticle, line 46: Found it here: https://ant.design/docs/react/customize-theme#consume-design-token

src/components/general/Typography/Typography.tsx
```diff
@@ -1,10 +1,18 @@
-import { Typography as AntTypography, type TypographyProps as AntTypographyProps } from 'antd'
+import {
+  Typography as AntTypography,
+  type TypographyProps as AntTypographyProps,
+  theme,
+  ConfigProvider as AntConfigProvider,
```
jared-dickman, line 5: probably we should import the ConfigProvider from the [aquaruim](https://github.com/mParticle/aquarium/blob/3b3344ac2ac4fb6ef0efa5bb90b4141b5b75f38d/src/components/other/ConfigProvider/ConfigProvider.tsx#L6-L5) and not ant, so that it applied the default theme that is exported from figma

jared-dickman: oh i didnt see the provider nesting below, nevermind 😄

jared-dickman (approved):

src/components/general/Typography/Typography.tsx
```diff
@@ -16,14 +24,42 @@ export const Typography = (props: ITypographyProps) => (
...
+
+  const fontSize = token[getFontSizeToken(size)]
+  const lineHeight = token[getLineHeightToken(size)]
+
+  return (
+    <ConfigProvider>
+      <AntConfigProvider theme={{ components: { Typography: { fontSize, lineHeight } } }}>
+        <AntTypography.Text {...props}>{props.children}</AntTypography.Text>
+      </AntConfigProvider>
+    </ConfigProvider>
```
tibuurcio, line 59: **thought** interesting use case cause we need to have 2 config providers. We might want to have a way to merge config providers in the future so that we only need to use one here. Maybe our `ConfigProvider` could provide as an export of the props it uses. Looks like currently it's only `theme={LightTheme}` and then we would be able to merge it somehow.

We could investigate this when we get to work on a better way of using the ConfigProvider.

tibuurcio (approved):
Only 1 thought on config providers, but looks good. Nice work.

src/components/general/Typography/Typography.tsx
ajenkins-mparticle: We don't need to have 2 config providers, the outer one in this component shouldn't exist, but that's a PR for a different time and not an easily made change. Well, actually it might be a simple change, but in order to understand if it's a simple change I need to find out everywhere where `Typography.Text` is used in aquarium consuming applications.

The second `ConfigProvider` is needed here - the theme [will be inherited from the top level one](https://ant.design/docs/react/customize-theme#nested-theme), no problem there - as it's the thing that sizes the Typography.

src/components/general/Typography/Typography.tsx
```diff
@@ -16,14 +20,38 @@ export const Typography = (props: ITypographyProps) => (
...
 export interface ITextProps extends AntTextProps {
+  size?: TypographySize
+}
+
+const getFontSize = (size: TypographySize): number => {
+  if (size === 'base') return 14
+  if (size === 'sm') return 12
+  if (size === 'lg') return 16
+  return 20
+}
```
tibuurcio, line 33: **suggestion** I get why we are hard coding the values since I'm involved in the thread in slack. Can I suggest we add some comments here explaining the reasoning behind the magic numbers (and the fact they should be temporary)😄

Makes the code be the source of truth without the need for outside context 👐🏼

ajenkins-mparticle: Absolutely, added as a TODO

src/components/general/Typography/Typography.tsx
jared-dickman: well luckily the typography component is still not used much in the aquarium or userland, i see 0 usages in indicative and cortex, and 3 in nancy 
`IncludeExcludeExpression.tsx`
`IncludeExcludeContainer.tsx`
`SetPredictionRange.tsx` [only in my current working branch]

src/components/general/Typography/Typography.tsx
```diff
@@ -16,13 +20,40 @@ export const Typography = (props: ITypographyProps) => (
...
+  size?: TypographySize
+}
+
+// TODO: Replace hardcoded values in getFontSize and getLineHeight with tokens when design is ready
+// These values are currently coming from https://www.figma.com/design/LffDbOUjeYqDMZ3djs9Cga/mParticle-Foundation-v1.0.1?node-id=3745-8164&m=dev
+const getFontSize = (size: TypographySize): number => {
+  if (size === 'base') return 14
+  if (size === 'sm') return 12
+  if (size === 'lg') return 16
+  return 20
```
jared-dickman, line 34: is this the same as `xl`, the default? if so thats a bit unclear, and would suggest making a default prop value like  [Icon.tsx](https://github.com/mParticle/aquarium/blob/a1bfb88721146c31ce3924ce887e9b936e67a335/src/components/general/Icon/Icon.tsx#L32-L31)

jared-dickman: oh its using a default prop inside the param ` size = 'base'` 🤔

src/components/general/Typography/Typography.tsx
```diff
@@ -16,13 +20,40 @@ export const Typography = (props: ITypographyProps) => (
...
+}
+
+const getLineHeight = (size: TypographySize): number => {
+  if (size === 'base') return 1.571428571428571
+  if (size === 'sm') return 1.666666666666667
+  if (size === 'lg') return 1.5
+  return 1.4
+}
+
+const Text = ({ size = 'base', ...props }: ITextProps) => {
```
jared-dickman, line 44: thinking id prefer to move the default props into `defaultProps` so the prop param can "do one thing" and just be a param 😄

ajenkins-mparticle: Do you mean like 

```
Text.defaultProps = {
  ...theDefaults
}
````?

ajenkins-mparticle: Oh yes, I can see from your comment above.

This is the preferred way to declare defaultProps for default values for functional components (not class components) and it has been for some time.

https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md
https://github.com/facebook/react/pull/16210

Does aquarium still prefer the old way?

ajenkins-mparticle: Actually, I just went to check react's codebase - it's not a preference anymore, it's a deprecation of support

https://github.com/search?q=repo%3Afacebook%2Freact%20%22Support%20for%20defaultProps%22&type=code

jared-dickman: damn i hadnt heard that! guess it makes sense to keep all prop-things together. can you create a TD ticket to update the icon and any other defaultProps

PR #263, opened 2024-06-05
feat: minimap active state

gabyzif:
## Summary
-  This PR fix the shadow hover effect
- Adds the active state

We agree with @gabrielmarquesf to not change the svg, so I applied some CSS to hide parts of the svg so the hover effects work 

After:
Screenshot 2024-06-05 at 9 40 10 PM
 
Screenshot 2024-06-06 at 10 00 35 PM

Before:
Screenshot 2024-06-05 at 9 32 44 PM
 
Screenshot 2024-06-06 at 9 59 36 PM

 
Screenshot 2024-06-06 at 10 01 23 PM

[Live](https://65aec861fa4adbf0708aebd0-hulqvugzdn.chromatic.com/?path=/story/aquarium-navigation-globalnavigation--mp)
## Testing Plan

- Closes https://go.mparticle.com/work/UNI-577


gabyzif:
Based on a convo w @gabrielmarquesf : even tho we love classes, i'll keep the css as before, hide the eliminated parts w css so we avoid css changes

src/components/navigation/MiniMap/SvgLinker.tsx
```diff
@@ -15,13 +15,18 @@ interface ISvgLinkerProps {
...
+
   const handleContainerClick = (e: React.MouseEvent) => {
     e.preventDefault()
     const target = e.target as HTMLElement
     const href = target.closest('a')?.getAttribute('href')
     const link = props.links.find(b => b.href === href)
 
-    if (link) props.onLinkClick(link)
+    if (link) {
+      !link.isUnauthorized && setActiveLink(link.href)
```
gabrielmarquesf, line 27: I am not sure if this will work in CDP, because I believe the MiniMap is been unmounted and mounted again when switching suites (would need to confirmate)
If that's true, you probably need to receive the `active` as a prop, probably as a field in the link obj

gabrielmarquesf: Also, when opening/refreshing the app using the full URL, you will not set the active suite.

gabyzif: ya, or I can handle this just by css

Seebiscuit: The GlobalNav already knows the current route. We don't have to couple to the Nav, but just like the Nav is getting the route, MiniMap should to. Because:
1. Consistency
2. One solution that works in all cases
3. Makes the Mini dumb which makes it easy to understand.

Seebiscuit: In case you're curious, [this is where the current route comes from](https://git.corp.mparticle.com/mParticle/mPServer/blob/98bfefc0cfe3842a88e8372080de3477a83753bd/src/UI/MP.UI.Web.Nancy/Assets/private/aurelia/src/layout/left-nav.ts#L12-L14), and it updates whenever the route changes.

src/components/navigation/MiniMap/miniMap.css
```diff
@@ -35,14 +40,30 @@
...
     fill: var(--mp-brand-secondary-7);
 }
+
+/* for the active state */
+.svg-linker-root__button--active > g > rect + rect   {
+    stroke: var(--mp-brand-primary-6);
+    stroke-width: var(--line-width-bold)
+}
+
+/* some fixes so the shadow on hover looks good without changing the svg */
```
Seebiscuit, line 63: **question** Is this going to break when we change the SVG?

src/components/navigation/MiniMap/minimap-svg.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 
 export const minimap = (
-  <svg width="385" height="374" viewBox="0 0 385 374" fill="none" xmlns="http://www.w3.org/2000/svg">
+  <svg width="385" height="374" viewBox="0 0 385 374" fill="none" xmlns="http://www.w3.org/2000/svg" className="u-overflow-visible">
```
Seebiscuit, line 4: **question** How are we going to remember to keep this here when we replace the SVG?

gabyzif: I can add it though css maybe its better

src/components/navigation/MiniMap/SvgLinker.tsx
```diff
@@ -15,29 +16,30 @@ interface ISvgLinkerProps {
...
-      !link.isUnauthorized && setActiveLink(link.href)
       props.onLinkClick(link)
     }
   }
 
-  return <div onClick={handleContainerClick}>{wrapButtonsIntoLinks(props.children)}</div>
+  return (
+    <div onClick={handleContainerClick} className="svg-linker__container">
+      {wrapButtonsIntoLinks(props.children)}
+    </div>
```
gabrielmarquesf, line 33: I like it 😄

src/components/navigation/MiniMap/miniMap.css
gabyzif: the overview map has the same structure as the minimap.. so it shouldn't

src/components/navigation/MiniMap/SvgLinker.tsx
```diff
@@ -15,29 +16,30 @@ interface ISvgLinkerProps {
...
+  )
 
   function wrapButtonsIntoLinks(parent: React.ReactNode): React.ReactNode {
     const wrapElement = (element: ReactElement): ReactElement => {
       const { id, children } = element.props
       const link = props.links.find(b => b.elementId === id)
 
       if (link) {
-        const isActiveClass = link.href === activeLink && ' svg-linker-root__button--active'
+        const isActiveClass = link.isActive && ' svg-linker-root__button--active'
```
gabrielmarquesf, line 42: you may render `false` in the className

src/components/navigation/MiniMap/miniMap.css
gabyzif: if they make edits on the minimap following the current svg, it should be good...

gabrielmarquesf (approved):

PR #265, opened 2024-06-06
chore: dont allow wrapping of query item text component

jared-dickman:
## Instructions

## Summary
- the typography component from the aquarium has `word-break: break-word;`, which causes the component to wrap (see screenshot)
this can be turned off by adding [white-space](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space): nowrap

## Screenshots
image

### DIFF
--- a/src/components/data-entry/QueryItem/Text.tsx
+++ b/src/components/data-entry/QueryItem/Text.tsx
@@ -7,7 +7,7 @@ export interface ITextProps {
 
 export const Text = ({ disabled = false, text }: ITextProps) => {
   return (
-    <Typography.Text disabled={disabled} style={{ textWrap: 'nowrap' }}>
+    <Typography.Text disabled={disabled} style={{ whiteSpace: 'nowrap' }}>
       {text}
     </Typography.Text>
   )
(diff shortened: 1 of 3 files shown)


src/components/data-entry/QueryItem/Text.tsx
```diff
@@ -6,7 +6,11 @@ export interface ITextProps {
 }
 
 export const Text = ({ disabled = false, text }: ITextProps) => {
-  return <Typography.Text disabled={disabled}>{text}</Typography.Text>;
+  return (
+    <Typography.Text disabled={disabled} style={{ whiteSpace: 'nowrap' }}>
```
tibuurcio, line 10: **question** This looks fine but what if we want it to wrap in a future usage? 🤔
Is there a way to override this with our current styling options?

jared-dickman: hm yea sure, would either have to pass in a new prop, or in userland probably wrap it in a div with a css class and target it that way

tibuurcio: Just noticed this is the specific QueryItem component. We should be fine.

tibuurcio (approved):

PR #266, opened 2024-06-07
fix: rename minimap links so it matches with the routing in Aurelia

gabyzif:
This PR changes the case of customer360 and dataPlatform to customer-360 and data-platform so it matches the routing in Nancy


tibuurcio (approved):
Although Aquarium should not care how things are defined in Nancy, The minimap is very much coupled with the suites implementation so it makes sense to do this change in the Aquarium in order to avoid having to do a transformation in Nancy before sending it here.

PR #267, opened 2024-06-08
chore: component -> query item - value select - number input 

jared-dickman:
## Instructions

## Summary
- extending the Query Item framework with a new Number types
- this has the advantage of including numerical support, decimals, scientific notation, removing letters, min/max, and more
- also includes changing the export type for nested components, so that its more difficult to use the component in a non-nested way

## Testing Plan
- storybook, new stories passing, will install this feature locally into nancy 

- Closes https://go.mparticle.com/work/UI3DM-1250

### DIFF
--- a/src/components/data-entry/QueryItem/Action.tsx
+++ b/src/components/data-entry/QueryItem/Action.tsx
@@ -8,7 +8,7 @@ export interface IActionProps {
   onClick?: () => void
 }
 
-export const Action = (props: IActionProps) => {
+const Action = (props: IActionProps) => {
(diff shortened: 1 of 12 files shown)


src/components/data-entry/QueryItem/Action.tsx
```diff
@@ -32,4 +32,6 @@ export const Action = (props: IActionProps) => {
       </Button>
     </>
   )
-}
\ No newline at end of file
+}
+
+export default Action
```
jared-dickman, line 37: ive moved these sub-components into default exports, so that they dont appear as options in the ide auto-import menu

PR #268, opened 2024-06-08
feat: query select number

jared-dickman:
## Instructions

## Summary
- extending the Query Item framework with a new Number types
- this has the advantage of including numerical support, decimals, scientific notation, removing letters, min/max, and more
- includes changing the export type for nested components, so that its more difficult to use the component in a non-nested way

- also adds `fix` and `chore` brances as allowable npm releases, because why not

## Testing Plan
- storybook, new stories passing, will install this feature locally into nancy 

- Closes https://go.mparticle.com/work/UI3DM-1250

### DIFF
--- a/release.config.cjs
+++ b/release.config.cjs
@@ -1,9 +1,22 @@
 module.exports = {
-  branches: ['main', {
-    name: 'feat/*',
-    channel: 'feature',
-    prerelease: "${name.split('/').slice(1).join('-').toLowerCase()}"
-  }],
+  branches: ['main', 
+    {
+      name: 'feat/*',
+      channel: 'feature',
+      prerelease: "${name.split('/').slice(1).join('-').toLowerCase()}"
+    },
+    {
+      name: 'chore/*',
+      channel: 'chore',
+      prerelease: "${name.split('/').slice(1).join('-').toLowerCase()}"
+    },
+    {
+      name: 'fix/*',
+      channel: 'fix',
+      prerelease: "${name.split('/').slice(1).join('-').toLowerCase()}"
+    },
+    
+  ],
   tagFormat: 'v${version}',
   repositoryUrl: 'https://github.com/mParticle/aquarium',
   plugins: [
--- a/src/components/data-entry/QueryItem/Action.tsx
+++ b/src/components/data-entry/QueryItem/Action.tsx
@@ -8,7 +8,7 @@ export interface IActionProps {
   onClick?: () => void
 }
 
-export const Action = (props: IActionProps) => {
+const Action = (props: IActionProps) => {
   let buttonClassNames: string = 'query-item query-item--action'
   if ((props.type ?? 'default') === 'default') buttonClassNames += ` query-item--secondary`
   if ((props.type ?? 'default') === 'disabled') buttonClassNames += ` query-item--disabled`
@@ -33,3 +33,5 @@ export const Action = (props: IActionProps) => {
     </>
   )
 }
+
+export default Action
--- a/src/components/data-entry/QueryItem/Action.tsx
+++ b/src/components/data-entry/QueryItem/Action.tsx
@@ -10,8 +10,9 @@ export interface IActionProps {
 
 const Action = (props: IActionProps) => {
   let buttonClassNames: string = 'query-item query-item--action'
-  if ((props.type ?? 'default') === 'default') buttonClassNames += ` query-item--secondary`
-  if ((props.type ?? 'default') === 'disabled') buttonClassNames += ` query-item--disabled`
+  const { type = 'default' } = props.type
+  if (type === 'default') buttonClassNames += ` query-item--secondary`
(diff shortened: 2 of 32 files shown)


release.config.cjs
```diff
@@ -1,9 +1,22 @@
...
+    {
+      name: 'chore/*',
+      channel: 'chore',
+      prerelease: "${name.split('/').slice(1).join('-').toLowerCase()}"
+    },
+    {
+      name: 'fix/*',
+      channel: 'fix',
+      prerelease: "${name.split('/').slice(1).join('-').toLowerCase()}"
+    },
```
tibuurcio, line 17: **praise** ✨ makes sense

src/components/data-entry/QueryItem/Action.tsx
```diff
@@ -8,7 +8,7 @@ export interface IActionProps {
   onClick?: () => void
 }
 
-export const Action = (props: IActionProps) => {
+const Action = (props: IActionProps) => {
   let buttonClassNames: string = 'query-item query-item--action'
   if ((props.type ?? 'default') === 'default') buttonClassNames += ` query-item--secondary`
   if ((props.type ?? 'default') === 'disabled') buttonClassNames += ` query-item--disabled`
```
tibuurcio, line 14: I know it's not from this PR but a quick suggestion if you think it's worth it.

```suggestion
  const { type = 'default' } = props.type
  if (type === 'default') buttonClassNames += ` query-item--secondary`
  if (type === 'disabled') buttonClassNames += ` query-item--disabled`
```

src/components/data-entry/QueryItem/NumberInput.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+  component: QueryItem.ValueSelector.NumberInput,
+  parameters: {
+    docs: {
+      description: {
+        component: 'This is the "Number Input" component of the QueryItem/ValueSelector component group.',
+      },
+    },
+  },
+  args: {
+    onChange: (value: string | number) => {
```
tibuurcio, line 15: Can't be a string since we are calling `parseFloat`
```suggestion
    onChange: (value: number) => {
```

src/components/data-entry/QueryItem/NumberInput.tsx
```diff
@@ -0,0 +1,42 @@
...
+  value?: number
+  disabled?: boolean
+  errorMessage?: string
+  placeholder?: string
+  min?: number
+  max?: number
+  onChange?: (value: number) => void
+}
+
+const NumberInput = (props: INumberInputProps) => {
```
tibuurcio, line 15: **thought** Interesting component cause it looks very much like the native InputNumber from Antd but encapsulates some styling from the query-item and adds an errorMessage prop. I wonder if it couldn't be the case that we allow those customizations in the native InputNumber component although don't think we should have a "default" error rendering there so it kinda makes sense for this to live in it's own shell.

src/components/data-entry/QueryItem/QueryItem.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react'
-import { ValueSelector } from 'src/components/data-entry/QueryItem/ValueSelector'
-import { Action } from './Action'
-import { Qualifier } from './Qualifier'
-import { Text } from './Text'
+import ValueSelector from 'src/components/data-entry/QueryItem/ValueSelector'
+import Action from 'src/components/data-entry/QueryItem/Action'
+import Text from 'src/components/data-entry/QueryItem/Text'
+import Qualifier from 'src/components/data-entry/QueryItem/Qualifier'
```
tibuurcio, line 5: **praise** nice change 👍🏼

src/components/data-entry/QueryItem/TextInput.tsx
```diff
@@ -29,8 +25,15 @@ export const TextInput = (props: ITextInputProps) => {
         className={inputClasses}
         value={props.value}
         placeholder={props.placeholder}
-        onChange={_onChange}></Input>
-      {props.errorMessage && <Typography.Text type="danger">{props.errorMessage}</Typography.Text>}
+        onChange={e => {
+          props.onChange?.(e.target.value)
+        }}
```
tibuurcio, line 30: **thought** Not from this PR but I don't think we should get rid of the event payload here and pass only the value down. What if we need it in the future? Although this is would be a breaking change.

tibuurcio (approved):
Couple of thoughts but otherwise looks good. Nice work.

src/components/data-entry/QueryItem/NumberInput.tsx
jared-dickman: yes maybe, also this follows the pattern we establisher for the Query Item and having "sub components" that specify more specific variants of the ant components for the query builder specifically

tibuurcio (approved):

tibuurcio (approved):

PR #272, opened 2024-06-10
chore: add new icons, update icon story to select a single icon

jared-dickman:
## Instructions

## Summary
- Add new icons from design
- add suffix Dt to DuoTone icons [BREAKING CHANGE]
- change story to support displaying a single icon 

## Testing Plan
- verify new icons in storybook

- Closes https://go.mparticle.com/work/PDOG-2109


src/components/general/Icon/Icon.tsx
```diff
@@ -16,13 +16,21 @@ export type IconColor =
   | 'brand'
 
 export interface IIconProps {
-  name: keyof typeof Icons
+  name//: if type==='duo' then name is keyof typeof DuoIcons, else if type==='light' then name is keyof typeof LightIcons
```
jared-dickman, line 19: **THIS IS A WIP**

jared-dickman: alternative idea from kevin, make the api more like ` <Icon name={ IconNames.prediction.duo } /> ` with no type prop

src/components/icons/index.ts
```diff
@@ -43,22 +44,25 @@ import TransformationsIcon from 'src/assets/svg/mp_pm_lt_transformations.svg?rea
 import UploadIcon from 'src/assets/svg/mp_act_it_upload.svg?react'
 import UserAttributeIcon from 'src/assets/svg/user-attribute.svg?react'
 import UserProfilesIcon from 'src/assets/svg/mp_pm_lt_userprofiles.svg?react'
-import UsersIcon from 'src/assets/svg/users.svg?react'
+import UsersIcon from 'src/assets/svg/mp_info_lt_users.svg?react'
```
jared-dickman, line 47: **todo:** caro says we should replace this old users icon [with padding on all sides] for the new one provided [with padding only on the top/bottom]

src/components/general/Icon/Icon.tsx
tibuurcio: Interesting 👀 I kinda like it.

src/components/icons/index.ts
jared-dickman: done

src/components/general/Icon/Icon.tsx
jared-dickman: great! how can we make it happen? perhaps pair program in it for a few minutes later today

nrmerlis (approved):

PR #282, opened 2024-06-12
Wrap radio group and button components

gabrielmarquesf:
## Instructions

## Summary

## Testing Plan

- [ x Was this tested locally? If not, explain why.


src/components/data-entry/Radio/Radio.tsx
```diff
@@ -14,5 +16,5 @@ export const Radio = (props: IRadioProps) => {
 
 // TODO Is there a way to type the props from Radio.Group better so that value types are inferred?
 // This happens with ant as well. <Radio.Group value={string} /> doesn't get properly propagated to the change event
```
tibuurcio, line 18: Yeah I've noticed that as well. I'm not 100% sure but Input.tsx might help here. We used some inspiration code from Antd's codebase there to be able to properly type the `ref` prop.

tibuurcio (approved):
Nice first PR! 🥳

PR #290, opened 2024-06-13
idea: remove wrapped but non-extended ant components

ajenkins-mparticle:
## Instructions

## Summary

- Removes all directly wrapped/non-extended components.
- Leave any components that had the slightest extension (wrapped in fragments, default props, etc)

## Testing Plan


jared-dickman:
im super curious about this. the idea was that we wanted to add a layer of separation between the aquarium and ant. incase we decide later to swap out ant in one or more components. 

can you briefly describe the guiding principles behind this pr and removing all wrapping?

jared-dickman:
this is def a great candidate to talk more through as a team! I thought more about it this weekend, and am not convinced. 

What is the principle here, to ship pre-configured components that are separated from ant, or to put configuration on the platform without a boundary between ant?

PR #291, opened 2024-06-14
chore: update husky to v9 and fix hooks

danielsiemens:
## Instructions

## Summary

- Updates Husky to latest version (9.0.11)
- Fixes an issue with `lint-staged` where running `eslint` and `stylelint` concurrently would cause deadlock/file permission issues
- Fixes stylelint errors by emptying `property-disallowed-list`
- Fixes minor complaint in .eslintrc.cjs that the file was not present in the ignorePatterns
- Prettier run on staged files
- `_variables.css` and `styles.css` regenerated from pre-commit hook running `npm run tokens-to-css`

## Testing Plan

- [] Was this tested locally? If not, explain why.

- Closes https://mparticle-eng.atlassian.net/browse/UNI-740


.eslintrc.cjs
```diff
@@ -1,132 +1,132 @@
...
+    '@typescript-eslint/promise-function-async': 'warn',
+    'react/react-in-jsx-scope': 'off',
+    'import/no-duplicates': 'off',
+    'react/jsx-boolean-value': 'warn',
+  },
+  globals: {
+    React: true,
+    expect: true,
+  },
+}
```
jared-dickman, line 132: :thisisfine:

src/styles/_variables.css
```diff
@@ -1,27 +1,26 @@
 :root {
   --blue: #1677ff;
-  --purple: #722ed1;
```
jared-dickman, line 3: one day we will agree on if these should be upper or lower case

that day will come in hundreds of years

package.json
```diff
@@ -89,14 +89,18 @@
...
+    "stylelint:staged": "stylelint --allow-empty-input --fix ",
     "tokens-to-css": "npx style-dictionary build --config ./style-dictionary.json"
   },
   "lint-staged": {
-    "*.{ts,tsx,css}": [
+    "*.{ts,tsx}": [
       "npm run lint:staged",
       "prettier --write"
     ],
+    "*.{css}": [
```
jared-dickman, line 101: wow, brilliant! 🎉

package.json
```diff
@@ -57,7 +57,7 @@
     "eslint-plugin-react-hooks": "4.6.0",
     "factory.ts": "1.4.1",
     "http-server": "14.1.1",
-    "husky": "8.0.3",
+    "husky": "^9.0.11",
```
jared-dickman, line 60: we've been pinning packages to specific versions, any reason to not do the same for husky?

jared-dickman (approved):

package.json
danielsiemens: ah i missed that! I guess i'll flip the question and ask is there a reason we're pinning versions?

danielsiemens: in any case pinned to 9.0.11!

jared-dickman (approved):

package.json
tibuurcio: I think no reason specifically but I've faced some problems in Nancy with packages auto-updating. Mostly aurelia stuff. That's how I learned the hard way not to get rid of Nancy's lock file. If we do, it breaks everything :harold:

I assume the decision comes from those sort of things.

tibuurcio:
Great first piece of work @danielsiemens!

PR #321, opened 2024-07-03
chore/more icons typed

tibuurcio:
# THIS WILL BE A BREAKING CHANGE
- todo:  will need to rename branch to `feat/`

## Summary
we will want to support having icons with the same name but a different type
for example, types DuoTone and Light are both used today in the platforms
but there are currently no cases of using one icon name with two types
we should support this because design expects it

## Testing Plan
testing done on story book [Icons](http://localhost:6006/?path=/docs/aquarium-general-icons--documentation) page 


ilyusha:
### Review Forge
[Click Here for AI Insights](https://jared-dickman.github.io/review-forge?prUrl=https://github.com/mParticle/aquarium/pull/321)

PR #325, opened 2024-07-08
feat: query item text props

jared-dickman:
# BREAKING CHANGE

## Summary
- align QueryItem.Text props with Typography (change `text` prop to `children`)
- add type prop to QueryItem.Text


tibuurcio (commented):
If we assign a default value for `type` do we make this not being a breaking change? Maybe worth it.

PR #326, opened 2024-07-11
Chore: Adding functionality to QueryItem.ValueSelector.NumberInput

nrmerlis:
## Summary

- INumberInputProps Interface is now being exported
- NumberInput validates if the received value is Not a Number. If it is, it will be set undefined to its current value.

## Testing Plan

- [x] Was this tested locally?
- This has been tested in Nancy codebase. It has been imported, and it is working as intended.

### DIFF
--- a/src/components/data-entry/QueryItem/NumberInput.tsx
+++ b/src/components/data-entry/QueryItem/NumberInput.tsx
@@ -2,7 +2,7 @@ import './query-item.css'
 import { InputNumber } from 'src/components'
 import { Typography } from 'src/components/general/Typography/Typography'
 
-interface INumberInputProps {
+export interface INumberInputProps {
   value?: number
   disabled?: boolean
   errorMessage?: string
--- a/src/components/data-entry/QueryItem/NumberInput.tsx
+++ b/src/components/data-entry/QueryItem/NumberInput.tsx
@@ -3,7 +3,7 @@ import { InputNumber } from 'src/components'
 import { Typography } from 'src/components/general/Typography/Typography'
 
 export interface INumberInputProps {
-  value?: number
+  value?: number | undefined
   disabled?: boolean
   errorMessage?: string
   autoFocus: boolean
@@ -21,6 +21,12 @@ const NumberInput = (props: INumberInputProps) => {
   let inputClasses = `query-item query-item--input-number`
   if (props.errorMessage) inputClasses += ' query-item--error'
 
+  const handleOnChange = (value: string | number | null | undefined) => {
+    const floatValue = parseFloat(value as string)
+    if (isNaN(floatValue)) return props.onChange?.(undefined)
+    props.onChange?.(floatValue)
+  }
+
   return (
     <>
       <InputNumber
@@ -34,10 +40,7 @@ const NumberInput = (props: INumberInputProps) => {
         min={props.min}
         step={props.step}
         onPressEnter={props.onPressEnter}
-        onChange={(value: string | number | null | undefined) => {
-          const floatValue = parseFloat(value as string) || undefined
-          props.onChange?.(floatValue)
-        }}
+        onChange={handleOnChange}
       />
 
       {props.errorMessage && <Typography.Text type="danger">{props.errorMessage}</Typography.Text>}
--- a/src/components/data-entry/QueryItem/NumberInput.tsx
+++ b/src/components/data-entry/QueryItem/NumberInput.tsx
@@ -21,11 +21,6 @@ const NumberInput = (props: INumberInputProps) => {
   let inputClasses = `query-item query-item--input-number`
   if (props.errorMessage) inputClasses += ' query-item--error'
 
-  const handleOnChange = (value: string | number | null | undefined) => {
-    const floatValue = parseFloat(value as string) || undefined
-    props.onChange?.(floatValue)
-  }
-
   return (
     <>
       <InputNumber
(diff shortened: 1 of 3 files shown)


src/components/data-entry/QueryItem/NumberInput.tsx
```diff
@@ -34,8 +34,8 @@ const NumberInput = (props: INumberInputProps) => {
         min={props.min}
         step={props.step}
         onPressEnter={props.onPressEnter}
-        onChange={(value: string | number | null) => {
-          props.onChange?.(parseFloat(value as string))
+        onChange={(value: string | number | null | undefined) => {
+          props.onChange?.(parseFloat(value as string) || undefined)
```
jared-dickman, line 38: **suggestion**
pull `parseFloat(value as string) || undefined` into a variable
also, i think the linter will suggest using `??` instead of `||`

src/components/data-entry/QueryItem/NumberInput.tsx
```diff
@@ -34,8 +34,8 @@ const NumberInput = (props: INumberInputProps) => {
...
         }}
```
tibuurcio, line 39: Hey @nrmerlis I have a few suggestions to improve this validation, but also I believe I bug can occur here if you're given the value of 0. `parseFloat('0') || undefined` will change the value to `undefined` because `parseFloat('0')` will become `0` which will be coerced to `false` when you would actually want the value to be set to 0 since it's a valid number.

Some other points of improvement I see:

- `parseFloat(value as string)` casting `as string` here doesn't actually do what we need since it only happens at compilation time.
- Using the coerced value of `parseFloat` is not a good idea because falsy values will return `NaN` which has a few problems, see [here](https://eslint.org/docs/latest/rules/use-isnan).
- If `value` can be undefined it would be interesting to reflect that in the interface as `value?: number | undefined`

My suggestion is to work a bit more on this validation by using `Number(value)` and `Number.isNaN` for those validations, without relying on coercing.

I'm not sure how easy it Aquarium's current setup but some tests for this would be great as well, if the task allows time for it.

tibuurcio (changes requested):
Left a suggestion on how to make the validation logic a bit less error prone 🙏🏼 
Let me know if it makes sense.

src/components/data-entry/QueryItem/NumberInput.tsx
nrmerlis: Imposible 🙃
image

tibuurcio: Yeah that's cause of the way `NaN` works... See my suggestions.

src/components/data-entry/QueryItem/NumberInput.tsx
nrmerlis: I've kept the parseFloat() method because Number(null) would return 0, and we want to be sure that if there is no number (null or undefined), we will get a NaN so then we can pass an undefined to the onChange fn.

nrmerlis: @tibuurcio, by the way, @testing-library/react is not installed on the project. Would you like me to create a branch so we can test react in the future?

tibuurcio: > I've kept the parseFloat() method because Number(null) would return 0, and we want to be sure that if there is no number (null or undefined), we will get a NaN so then we can pass an undefined to the onChange fn.

Sounds good @nrmerlis! It's bug-free now 🙏🏼 

Have one last suggestion to make it clearer about the early return, conceptually works the same but I don't it's a good practice to return on `void` fns (is it? I might be wrong).

```ts
const handleOnChange = (value: string | number | null | undefined) => {
  const floatValue = parseFloat(value as string)
  if (isNaN(floatValue)) {
    props.onChange?.(undefined)
    return 
  }
  props.onChange?.(floatValue)
}
```

Other than that, you shouldn't need the casting of `(value as string)`... Is TS complaining with you for `parseFloat`?

tibuurcio: > @tibuurcio, by the way, @testing-library/react is not installed on the project. Would you like me to create a branch so we can test react in the future?

We haven't settled on the testing strategies for the Aquarium but would be nice to have a POC using testing-library so we could see how would it fit in the library. If you're able to create it that would be nice @nrmerlis!

nrmerlis: > > I've kept the parseFloat() method because Number(null) would return 0, and we want to be sure that if there is no number (null or undefined), we will get a NaN so then we can pass an undefined to the onChange fn.
> 
> Sounds good @nrmerlis! It's bug-free now 🙏🏼
> 
> Have one last suggestion to make it clearer about the early return, conceptually works the same but I don't think it's a good practice to return on `void` fns (is it? I might be wrong, never seen it though).
> 
> ```ts
> const handleOnChange = (value: string | number | null | undefined) => {
>   const floatValue = parseFloat(value as string)
>   if (isNaN(floatValue)) {
>     props.onChange?.(undefined)
>     return 
>   }
>   props.onChange?.(floatValue)
> }
> ```
> 
> Other than that, you shouldn't need the casting of `(value as string)`... Is TS complaining with you for `parseFloat`?

Yeah, if I the value is not casted it complains! 

```
TS2345: Argument of type string | number | null | undefined is not assignable to parameter of type string
Type undefined is not assignable to type string
```

Secondly, about the void return, I've changed the function like this So that there is no doubt  :)

```
const handleOnChange = (value: string | number | null | undefined) => {
    const floatValue = parseFloat(value as string)
    isNaN(floatValue) ? props.onChange?.(undefined) : props.onChange?.(floatValue)
  }
```

PR #340, opened 2024-07-24
feat: add lock/unlock icons; rename previous lock to paywall

nastyastavitskaya:
## Summary

- Adding new lock/unlock icons.
- Renaming previos `lock` icon to `paywall`
https://mparticle-eng.atlassian.net/browse/UNI-817

## Testing Plan
Screenshot 2024-07-24 at 9 55 30 AM


ilyusha:
### Review Forge
[Click Here for AI Insights](https://jared-dickman.github.io/review-forge?prUrl=https://github.com/mParticle/aquarium/pull/340)

src/components/icons/index.ts
```diff
@@ -27,7 +27,7 @@ import IdentityIcon from 'src/assets/svg/mp_pm_lt_identity.svg?react'
 import JumpToIcon from 'src/assets/svg/mp_act_lt_jump-to.svg?react'
 import LightBulbIcon from 'src/assets/svg/lightbulb.svg?react'
 import LiveStreamIcon from 'src/assets/svg/mp_pm_lt_live-stream.svg?react'
-import LockIcon from 'src/assets/svg/lock.svg?react'
+import PaywallIcon from 'src/assets/svg/paywall.svg?react'
```
nastyastavitskaya, line 30: @jared-dickman hey Jared, this is the new name for the lock icon that was used in Analytics

gabyzif (approved):

src/components/icons/index.ts
tibuurcio, line 30: Can we keep both names available and add a `@deprecated` notice somehow to the old one as to not introduce the breaking change in a single release? This way we have some time to update in Analytics. It's a pattern we talked about, I think it's a good case for introducing it.

nastyastavitskaya: @tibuurcio I'm checking with Jared, cause I can't fine it in indicative code-base

nastyastavitskaya: ok, it's not breaking change for analytics! since the component that uses this icon is in aquarium and I updated it in the file below

PR #355, opened 2024-08-05
feat: add variants to icons

gabyzif:
2. PR title prefix should state semantic value for the change, usually `feat:`, `fix:` or `chore:`. [Source.](https://github.com/mParticle/mparticle-workflows/blob/main/.github/workflows/pr-title-check.yml)
3. PR branch prefix should state the same as the above with a `/`, usually `feat/`, `fix/` or `chore/`. [Source.](https://github.com/mParticle/mparticle-workflows/blob/main/.github/workflows/pr-branch-check-name.yml)

## Summary

The purpose of this PR is adding a variant in the icon component and making a better DX experience on adding the icons.

Design wants to organize the icons in storybook by categories on the future so I thought having an dictorioary of icons would make it easier for this case (or if we need other properties)

I added a documentation in the icons stories where it might be good to have all the steps we need to do when adding one.

I also started working on adding a @deprecated for the names of the icons.  
## Testing Plan


tibuurcio:
Summary is telling me the "what" changes but I can get that from looking at the files. I usually use it to understand the "why" we're doing the change. Context that's not on the code.

src/constants/Icons.ts
```diff
@@ -1,129 +1,320 @@
...
+import NextIcon from 'src/assets/svg/mp_act_lt_next.svg?react'
+import OpenTabIcon from 'src/assets/svg/mp_act_lt_open-tab.svg?react'
+import PreviousIcon from 'src/assets/svg/mp_act_lt_previous.svg?react'
+import DirectoryIcon from 'src/assets/svg/mp_pm_lt_directory.svg?react'
+import LockIcon from 'src/assets/svg/mp_act_lt_lock.svg?react'
+import UnlockIcon from 'src/assets/svg/mp_act_lt_unlock.svg?react'
+import NotificationIcon from 'src/assets/svg/mp_pm_lt_notification.svg?react'
+import PremiumIconDt from 'src/assets/svg/mp_info_dt_premium.svg?react'
+
+export const Icons: Record<IconNames, IconOptions> = {
```
tibuurcio, line 66: ✨ praise for nice typing

src/types/iconTypes.ts
```diff
@@ -0,0 +1,84 @@
...
+export type IconVariant = 'light' | 'duo-tone'
+
+export type IconOptions = RequireAtLeastOne<
+  {
+    light?: React.ComponentType<React.SVGProps<SVGSVGElement>>
+    'duo-tone'?: React.ComponentType<React.SVGProps<SVGSVGElement>>
+    default: IconVariant
+    deprecated?: boolean
+  },
+  'light' | 'duo-tone'
```
tibuurcio, line 12: This should work

```suggestion
  IconVariant
```

tibuurcio (commented):
Nice work, left some suggestions!

src/components/general/Icon/Icon.tsx
```diff
@@ -16,17 +18,29 @@ export type IconColor =
...
+  variant?: IconVariant
 }
 
-export const Icon = (props: IIconProps) => {
-  const { color = 'default', size = 'lg' } = props
+export const Icon: React.FC<IIconProps> = ({ name, color = 'default', size = 'lg', variant }) => {
+  const iconVariants = Icons[name] as IconOptions
+
+  if (iconVariants?.deprecated) {
+    console.warn(`Icon with name "${name}" is deprecated. Please use "predictions" instead.`)
```
tibuurcio, line 31: I like the warning, but we can't fix "predictions" here 😅 

What if the `deprecated` type was not a `boolean` but instead a `string` with the value of the new icon that should be used? Does it make sense? 🤔

gabyzif: yes 🎉

src/components/general/Icon/Icon.tsx
```diff
@@ -1,4 +1,6 @@
+import React from 'react'
 import { Icons } from 'src/constants/Icons'
+import { IconOptions, IconVariant, IconNames } from 'src/types/iconTypes'
```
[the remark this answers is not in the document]
gabyzif: yup, but I think eslint was complaing about it

src/components/general/Icon/Icon.tsx
```diff
@@ -16,17 +18,29 @@ export type IconColor =
...
+  name: IconNames
   color?: IconColor
   size?: IconSize
+  variant?: IconVariant
 }
 
-export const Icon = (props: IIconProps) => {
-  const { color = 'default', size = 'lg' } = props
+export const Icon: React.FC<IIconProps> = ({ name, color = 'default', size = 'lg', variant }) => {
+  const iconVariants: IconOptions = Icons[name]
```
tibuurcio, line 28: **suggestion** better naming here, this is not an Icon variant, it is an icon and variant is a property of it. How do you feel about using `icon`?

src/types/icons.ts
```diff
@@ -0,0 +1,84 @@
...
+    default: IconVariant
+    deprecated?: string
+  },
+  IconVariant
+>
+
+/**
+ * @deprecated Use "predictions" instead.
+ */
+export type DeprecatedIconNames = 'sparkles'
```
tibuurcio, line 18: This is currently not working as expected, right? Can we add that to this comment and maybe mentioning we should try to find a better way in the future?

tibuurcio (commented):
Mostly good to go, left 2 comments.

src/components/general/Icon/Icon.stories.tsx
```diff
@@ -13,24 +14,48 @@ export const IconTable: React.FC<IIconProps> = ({ color = 'black', size = 'lg',
...
+2. **Update Icon Constants**: Update the \`Icons\` object in \`src/constants/Icons\` to include the new icons.
+3. **Use Icons**: Use the updated icons in your components by referencing their names.
+
+#### Example Usage
+\`\`\`jsx
+<Icon name="new-icon" size="lg" color="primary" variant="light" />
+\`\`\`
+
+This will render the new icon with the specified size, color, and variant.
+`
```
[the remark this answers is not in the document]
gabyzif: this works btw

src/components/icons/index.ts
```diff
@@ -123,5 +123,5 @@ export {
   LockIcon,
   UnlockIcon,
   NotificationIcon,
-  PremiumDtIcon,
+  PremiumIconDt,
```
nastyastavitskaya, line 126: if the names are changed, does it mean they will get broken in the apps?

nastyastavitskaya (approved):

PR #359, opened 2024-08-06
fix: make the workspace-selector label smaller

gabyzif:
## Summary

This PR makes the workspace selector label to match the size of the rest of the nav labels.

## Testing Plan

- Closes https://mparticle-eng.atlassian.net/browse/UNI-831


src/components/navigation/GlobalNavigation/global-navigation.css
```diff
@@ -251,7 +252,7 @@
 
 .globalNavigation__iconLabel {
   color: var(--color-text);
-  font-size: 10px;
+  font-size: var(--label-size);
```
tibuurcio, line 255: How is it smaller if the variable has the same value as before? (10px). I'm confused 😅

nastyastavitskaya: this is the correct value, Gaby just replaced it with a variable. the workspace selector text was not correct and was suppose to match this font-size

tibuurcio (approved):

src/components/navigation/GlobalNavigation/global-navigation.css
```diff
@@ -94,7 +95,7 @@
 }
 
 .globalNavigation__item.workspaceSelector__menuItem{
-  height: var(--nav-item-height-large) !important;
+  height: calc(var(--control-height-lg) * 2) !important;
```
nastyastavitskaya, line 98: can we remove `var(--nav-item-height-large)` if it's no longer used?

nastyastavitskaya (approved):
looks so much better! 👏 🎉

PR #363, opened 2024-08-07
fix: icon color not changing on button hover

nastyastavitskaya:
## Summary

- Icon component used `default` as color. It led to an issue when the icon was used in the button the color update on the hover state was not reflected in the icon. This PR brings an update to the icon with an `inherit` color class, and to the button to set icons within the button component to use the `inherit ' color by default.

## Testing Plan

- Button stories were updated with icons `inherit` and `default` colors.
- Global nav: icon in the search button is now changing color on hover.

- Closes https://mparticle-eng.atlassian.net/browse/UNI-562

##  Screenshots
<details><summary>Nav</summary>Screenshot 2024-08-07 at 4 31 13 PM</details>

<details><summary>Button with default and inherit icon colors</summary>Screenshot 2024-08-07 at 4 34 27 PM</details>


gabyzif (approved):

src/components/general/Button/Button.tsx
```diff
@@ -9,15 +12,26 @@ export interface IButtonProps extends AntButtonProps {
...
 }
 export const Button = (props: IButtonProps) => {
   const classMap = {
     'with-new-icon': 'u-display-flex u-align-items-center u-justify-center',
   }
 
+  const buttonIcon =
+    React.isValidElement<IIconProps>(props.icon) && props.icon.type === Icon
+      ? React.cloneElement(props.icon, {
+          color: props.icon.props.color ?? 'inherit',
```
jared-dickman, line 25: :blinking:

nastyastavitskaya: tell me there's a better way to do it :))

jared-dickman: should we pass in the other icon props as a spread??

nastyastavitskaya: true...but I mean the entire approach of setting the default color for icon in button. I'm unsure about it. I've asked gpt, but I'm not trusting it 🤖

src/components/general/Button/Button.tsx
```diff
@@ -9,15 +13,26 @@ export interface IButtonProps extends AntButtonProps {
...
   const classMap = {
     'with-new-icon': 'u-display-flex u-align-items-center u-justify-center',
   }
 
+  const buttonIcon =
+    React.isValidElement<IIconProps>(props.icon) && props.icon.type === Icon ? (
+      <Icon {...props.icon.props} color={props.icon.props.color ?? 'inherit'} />
+    ) : (
+      props.icon
+    )
```
tibuurcio, line 28: Interesting! Neven seen this.

tibuurcio: Where did you get the idea for this part? `props.icon.type === Icon`

tibuurcio (approved):
LGTM

src/components/general/Button/Button.tsx
nastyastavitskaya: me neither! I initially just passed the color='inherit' in the search button icon, but I didn't want to do it every time we use the icon in the button. So I asked chat gpt, what is the best way to have an Icon component with one default color, but set the color to another value if this component is used within another component as a prop. it gave me this

tibuurcio: Yeah this is super interesting! Don't know what magic it does but it works. Did some testing locally printing some consoles and works wonders. Nice!

image

tibuurcio (approved):

PR #367, opened 2024-08-12
feat: update SuiteLogo in GlobalNav to render nav switcher Tour

nastyastavitskaya:
## Summary

- Based on the user's request, we will move the nav switcher (minimap) to the top of the nav bar and open it on suite logo hover. To bring awareness to the change, we are adding a Tour component, to highlight the area of change. This PR contains the implementation of this component.

## Testing Plan

- Run the `MP With Nav Switcher Tour` story in GlobalNavigation. Clicking on the SuiteLogo to see the tour popup.
Screenshot 2024-08-12 at 3 32 17 PM

- Closes https://mparticle-eng.atlassian.net/browse/UNI-881

### DIFF
--- a/src/components/navigation/GlobalNavigation/SuiteLogo.tsx
+++ b/src/components/navigation/GlobalNavigation/SuiteLogo.tsx
@@ -24,23 +24,11 @@ export function SuiteLogo({
 }: IGlobalNavigationLogo) {
   const logoRef = useRef(null)
 
-  const navSwitcherStep: ITourProps['steps'] = [
-    {
-      title: 'Navigate mParticle effortlessly!',
-      description: 'Switch between product suites anytime using this selector.',
-      placement: 'right',
-      target: () => logoRef.current,
-      nextButtonProps: {
-        children: 'Close',
-      },
-    },
-  ]
-
   return (
     <>
       <div ref={logoRef}>
         {renderNavLogo()}
-        {navSwitcherTourOptions && renderNavTour(navSwitcherStep, navSwitcherTourOptions)}
+        {navSwitcherTourOptions && renderNavTour(navSwitcherTourOptions)}
       </div>
     </>
   )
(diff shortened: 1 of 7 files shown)


ilyusha:
### Review Forge
[Click Here for AI Insights](https://supreme-adventure-m612l2p.pages.github.io?prUrl=https://github.com/mParticle/aquarium/pull/367)

src/components/navigation/GlobalNavigation/SuiteLogo.tsx
```diff
@@ -12,30 +15,73 @@ function isStringIcon(icon: ReactNode | string): icon is keyof typeof Icons {
...
+    )
   }
 
-  return (
-    <Center vertical className="globalNavigation__suiteLogo" onClick={onSuiteLogoClick}>
-      <NavigationIcon icon={getIcon()} label="" hideLabel className={classMap[type]} />
-      {label}
-    </Center>
-  )
+  function renderNavTour(props: INavSwitcherTourOptions) {
```
tibuurcio, line 64: I'm always struggling to know how much surface for controlling components we should give to our users when it comes to the leftnav. 

On one hand, we've had instances where we had to release new Aquarium versions only for adding missing props. In this case, it would be good to make `INavSwitcherTourOptions` accept more things like the rest of `Tour` props and do a `{...props}` at it.

On the other hand, we do want consistency across platforms since it's supposed to be the same nav...

So for this case I have 2 suggestions:

- We do not want to allow our users to customize the `steps`, but the contents of it might need variation between platforms. Suggestion is to add optional `title` and `description` values to the `INavSwitcherTourOptions` so that we can customize it if needed.
- Other `Tour` component props should be fine for us to give access in userland and not need to add it in future versions, so suggestion is to make `INavSwitcherTourOptions` extend TourProps and do a spread.

How does it those sound?

nastyastavitskaya: > I'm always struggling to know how much surface for controlling components we should give to our users when it comes to the leftnav.

it was my biggest struggle! I was trying to take inspiration from the new UI notification popover since they both are throw-away code (this Tour is just to bring awareness of the minimap moved to the top, and will be replaced with userpilot). but your suggestions make a perfect sense, thank you 🙌

tibuurcio (approved):

PR #370, opened 2024-08-14
fix: improve workspace selector performace

gabrielmarquesf:
## Summary
Some users have a long list of orgs/accounts/workspaces available, which was causing some slowness on this component, because it was rendering the whole list of items on every re-render.
This PR changes it to use virtualization, which means that it will render only the items visible to the user. 

## Testing Plan

- used the `mp-thousand-orgs` test in storybook to confirm performace improvments
- validated no side effects on other use-cases

- Closes https://go.mparticle.com/work/UNI-830


ilyusha:
### Review Forge
[Click Here for AI Insights](https://supreme-adventure-m612l2p.pages.github.io?prUrl=https://github.com/mParticle/aquarium/pull/370)

tibuurcio (approved):
Reviewed sync as PR buddy 🔥

PR #376, opened 2024-08-21
fix: update nav link active background color

nastyastavitskaya:
## Summary

- We used the wrong color token for an active link in the nav bar, and with the recent tokens update it looks a bit odd. Thie PR brings an update to use the right color token.

## Testing Plan

- Open global nav stories. Confirm that the active link is using the correct background-color (`var(--control-item-bg-active)`)


ilyusha:
### Review Forge
[Click Here for AI Insights](https://supreme-adventure-m612l2p.pages.github.io?prUrl=https://github.com/mParticle/aquarium/pull/376)

gabrielmarquesf (approved):

PR #422, opened 2024-09-20
fix: notification center export z-index

leobel:
## Summary

This PR will fix exporting `NotificationCenterZIndex`

## Testing Plan

- Tested with a release from the branch in CDP and Analytics

- Closes https://go.mparticle.com/work/SQDUI-2305

### DIFF
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -109,8 +109,8 @@ export type {
   IWorkspaceSelectorDisplayItem,
   IWorkspaceSelectorItem,
 } from './navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelectorItems'
-export type {
-  INotificationCenterProps,
+export {
+  type INotificationCenterProps,
   NotificationCenterZIndex,
 } from './navigation/GlobalNavigation/NotificationCenter'
 export { SuiteLogo } from './navigation/GlobalNavigation/SuiteLogo'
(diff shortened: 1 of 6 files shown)


ilyusha:
### Review Forge
[Click Here for AI Insights](https://supreme-adventure-m612l2p.pages.github.io?prUrl=https://github.com/mParticle/aquarium/pull/422)

src/components/index.ts
```diff
@@ -109,8 +109,8 @@ export type {
   IWorkspaceSelectorDisplayItem,
   IWorkspaceSelectorItem,
 } from './navigation/GlobalNavigation/WorkspaceSelector/WorkspaceSelectorItems'
-export type {
-  INotificationCenterProps,
+export {
+  type INotificationCenterProps,
```
tibuurcio, line 113: This feels weird, `type` is only applying to `INotificationCenterProps` here and seems like `NotificationCenterZIndex` was already being exported. @leobel were you able to validate this in the platforms with the feature version?

tibuurcio: Oh I think I got it. Meant weird not because of the usage of type but because the PR talks about fixing the z-index export but at first glance it is changing the `INotificationCenterProps`. But it's actually changing both.

So before, both `INotificationCenterProps` and `NotificationCenterZIndex` were being exported as types. This PR is changing **both of them**, NotificationCenterZIndex is now exported as a value and only INotificationCenterProps is kept as a type. That's the correct behaviour right? That's hard to understand at first sorry about the confusion!

With that, this is safe for approval 👍🏼

tibuurcio (approved):

gabyzif (approved):

PR #433, opened 2024-10-03
chore: fixing antd and dayjs peer dependencies

tibuurcio:
## Summary

We've had some problems recently related to library versions when trying to upgrade Antd version in the platforms and also when linking Aquarium locally for testing purposes.

This PR fixes both Antd and dayjs library versions so that we're always sure to be dealing with fixed versions.

## Testing Plan

The goal here is to release a feature version and perform the following tests:
- [X] Test CDP using Antd `5.20.0` and `5.20.6`
- [X] Test Analytics using Antd using Antd `5.20.0` and `5.20.6`
- [X] Test the local linking process in CDP
- [X] Test the local linking process in Analytics

- Closes https://go.mparticle.com/work/UNI-1098


ilyusha:
### Review Forge
[Click Here for AI Insights](https://supreme-adventure-m612l2p.pages.github.io?prUrl=https://github.com/mParticle/aquarium/pull/433)

package.json
```diff
@@ -24,9 +24,10 @@
     "dist/style.ts"
   ],
   "peerDependencies": {
-    "antd": ">=5.20.0",
+    "antd": "5.20.x",
```
danielsiemens, line 27: I've actually not seen this notation before with `x`. Is there a reason not to specify ^5.20.0?

danielsiemens (approved):

package.json
tibuurcio: GPT taught me. That's actually a great callout, I need to double check it's not hallucinating 😅 

Initially I thought let's go with a fixed version, but then patches are good to have. I believe `^5.20.0` will allow minor bumps as well, which I don't think we want for Antd cause they change a bunch of stuff in their minors...

danielsiemens: If we wanted to keep the version fixed i see no reason why that would cause an issue (downstream repos that update their aquarium would immediately see npm warnings that a peer dependency is not met and update accordingly).

I actually found that NPM defers to semver for syntax, which pointed me to this [helpful link](https://github.com/npm/node-semver?tab=readme-ov-file#advanced-range-syntax). So if we wanted to allow for patch versions, i believe your code is valid, or we could do:

`"antd": "~5.20.0",`

I'm happy with either personally. As long as the `x` works 😄

tibuurcio: Yeha perfect, that's what I want: `1.2.x := >=1.2.0 <1.3.0-0 (Matching major and minor versions)`

PR #470, opened 2024-10-22
docs: moves components to not-prod-ready folder and creates docs file…

tibuurcio:
## Summary

This PR restructures current documentation per Design team's request. It essencially:
- Creates the "Not Prod Ready" category and move some components there
- Creates a Documentation.mdx file for all `Data Display` and `Data Entry` components so that Design can work on the contents of it without having to mess with Stories.
- Restructure some other categories like moving "Other" to "Contributing/Tooling" and moving "Component Process" under "About"

This also removes the [root Antd `App`](https://ant.design/components/app) component since we were not using it and I feel like we shouldn't. It might be useful in the future for handling global Modal stuff but let's get there when we need to. I've verified our platforms are not using it. 

## Testing Plan

- Closes https://go.mparticle.com/work/UNI-1144


ilyusha:
### Review Forge
[Click Here for AI Insights](https://supreme-adventure-m612l2p.pages.github.io?prUrl=https://github.com/mParticle/aquarium/pull/470)

tsconfig.json
tibuurcio, line 1: This change is also present in my other Table Docs PR and it's needed when running the project through VSCode. Not sure why Rider never complaing, also doesn't seem like "bundler" is a valid option here so not sure where did it came from. Either way this should be `node`

gabyzif (approved):

PR #471, opened 2024-10-23
chore: update storybook

gabyzif:
## Summary
This PR updates Storybook and it's dependencies

## Testing Plan

- Closes https://go.mparticle.com/work/UNI-1021


.storybook/main.ts
```diff
@@ -30,10 +30,10 @@ const config: StorybookConfig & StorybookVitePlugins = {
         },
       },
     },
+    '@chromatic-com/storybook',
```
tibuurcio, line 33: Why do we need chromatic? I was thinking we would get rid of it now that we have PR envs?

gabyzif: I was thinking the same thing

.storybook/main.ts
```diff
@@ -49,6 +48,10 @@ const config: StorybookConfig & StorybookVitePlugins = {
   core: {
     disableTelemetry: true, // requested by security
   },
+
+  typescript: {
+    reactDocgen: 'react-docgen-typescript',
+  },
```
tibuurcio, line 54: Why is that needed?

.storybook/preview.tsx
```diff
@@ -52,6 +52,8 @@ const preview: Preview = {
       },
     },
   },
+
+  tags: ['autodocs'],
```
tibuurcio, line 56: same as the above

gabyzif: This happened when upgrading SB

Before we we had the autodocs placed on main here:

Screenshot 2024-10-24 at 5 37 53 PM

when I was upgrading SB this promt:

Screenshot 2024-10-24 at 5 24 24 PM

so basically it did the migration to the new version automatically.


more info here: [migration docs](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#autodocs-changes )

.storybook/main.ts
gabyzif: When doing the migration this was prompt:

Screenshot 2024-10-24 at 5 45 54 PM

So basically SB swiched to react-docgen but we had react-docgen-typescript on place and by doing this we are keeping it and not moving to react-docgen.

[More info here](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#react-docgen-component-analysis-by-default 
)

PR #472, opened 2024-10-23
refactor: update SuiteSelectorSuiteLink with a and href to handle right-click

nastyastavitskaya:
## Summary

- To enhance UX, the links in the nav switcher are updated with `a` and href to handle opening in a new tab with the right-click.
## Testing Plan

- Run global navigation story with nav switcher. Confirm the right-click is now supporting an option to open link in a new tab.
Screenshot 2024-10-23 at 1 51 12 PM

- Closes part of https://mparticle-eng.atlassian.net/browse/UNI-1146


ilyusha:
### Review Forge
[Click Here for AI Insights](https://supreme-adventure-m612l2p.pages.github.io?prUrl=https://github.com/mParticle/aquarium/pull/472)

gabyzif:
Looking great! However, I think we should prevent right-click functionality for elements in a disabled state.

tibuurcio (approved):
lgtm! :)

nastyastavitskaya:
> > Looking great! However, I think we should prevent right-click functionality for the elements in a disabled state.
> 
> That's an interesting one... right-click is a native feature so I wouldn't prevent it but you're right @gabyzif we might want to remove the `<a>` if it's disabled 🤔

good point. I don't really like to mess up by rendering different html tags. I think the easiest would be to set the CSS rule `pointer-events: none;`.  the downside for it is if we want any tooltip or something it won't work. but I think it should not be the case for this component

tibuurcio (approved):

PR #486, opened 2024-11-05
chore: better select prop types to match antd exports

tibuurcio:
## Summary

This PR makes Aquarium Select component match the way Antd exports its props so we are able to use generics while developing with the Select component.

It also fixes the `suffixIcon` to be the `dropdownOpen` icon from Aquarium.

## Testing Plan

- Closes https://go.mparticle.com/work/UNI-1075


ilyusha:
### Review Forge
[Click Here for AI Insights](https://supreme-adventure-m612l2p.pages.github.io?prUrl=https://github.com/mParticle/aquarium/pull/486)

ajenkins-mparticle (approved):
This is great, but the name of this PR hides the fact that you've defaulted the `suffixIcon` for the `Select` to something specific. It's a good thing for consistency, but probably should add a note about that.

tibuurcio:
> This is great, but the name of this PR hides the fact that you've defaulted the `suffixIcon` for the `Select` to something specific. It's a good thing for consistency, but probably should add a note about that.

Oh definitely, thanks for pointing that out. I've been working on this since last week and totally forgot I had done that.

src/components/data-entry/QueryItem/Qualifier.tsx
```diff
@@ -21,6 +21,8 @@ const Qualifier = (props: IQueryItemQualifierProps) => {
     defaultValue: props.options?.length ? props.options[0].value : undefined,
     menuItemSelectedIcon: node =>
       node.isSelected ? <CheckIcon className="query-item-qualifier__item-selected-icon" /> : null,
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error Introduced when we properly typed the Select value and option generics from Aquarium. Need to double check to fix this.
```
tibuurcio, line 25: @jared-dickman @ZoeAstra would any of you guys be able to take a quick look at this? I didn't want to change anything without the proper context. Ideally the `<Select>` used here in Qualifier would be typed and hopefully this problem would be avoided 🙌🏼

PR #491, opened 2024-11-14
chore: swap default expand icon for collapse component

tibuurcio:
## Summary

This PR sets the default icons for the Collapse component usage and fix table advanced filters docs to use it.

## Testing Plan


gabyzif (approved):

src/components/data-display/Collapse/Collapse.tsx
```diff
@@ -1,13 +1,18 @@
...
 
 export interface ICollapseProps extends AntCollapseProps {}
 
 export const Collapse = (props: ICollapseProps) => {
   return (
     <ConfigProvider>
-      <AntCollapse {...props} />
+      <AntCollapse
+        expandIcon={({ isActive }): React.JSX.Element =>
+          isActive ? <Icon name="dropdownOpen" size="sm" /> : <Icon name="next" size="sm" />
```
nastyastavitskaya, line 12: why does the native ant icon not work for us?

nastyastavitskaya (approved):

src/components/data-display/Collapse/Collapse.tsx
tibuurcio: It's a different chevron, same as the Select component. Our chevron came from Design and looks better IMO
image
image

nastyastavitskaya: it does look better indeed

nastyastavitskaya (approved):

PR #514, opened 2024-12-16
Foundations

soniakose:
Test for hex in color palette descriptions


tibuurcio:
Merging this without approvals since it's docs copy changes coming directly from the Design team

PR #516, opened 2024-12-16
docs: typography

soniakose:
Typography doc updates


tibuurcio:
Merging this without approvals since it's docs copy changes coming directly from the Design team

PR #530, opened 2025-01-17
fix: double scroll on Table

danielsiemens:
## Summary

- Creates a new story for a Table with Fixed Header/Sticky Columns (to mimic functionality where this issue was observed)
- Adds CSS to hide scrollbars when `sticky` is `true`:
  - Hides the horizontal scroll on the table itself
  - Hides an additional horizontal scrollbar that would only show on vertical scroll

## Testing Plan

Tested and validated on Chrome, Firefox, Safari

- Closes https://mparticle-eng.atlassian.net/browse/UNI-1299


tibuurcio (approved):
Nice, tested on PR env. `className` override also looks good.

PR #532, opened 2025-01-17
docs: documentation pages now appear first when clicking a story

tibuurcio:
## Summary

The goal of this PR is to make our `Documentation` pages be the first ones in the list so that when a user clicks in a Story it goes directly to them instead of an example story. Storybook 7 provides 2 ways of sorting stories: 1. providing a nested `options.storySort.order` array containing stories titles in sorted order or 2. providing an `options.storySort` function that acts as a standard sort function (receives 2 stories, returns 1, 0, -1). 

I went down the rabbit hole of implementing our own sorting function because I couldn't figure it out why our documentation pages were not being considered until I've stumbled upon `options.storySort.includeNames` option which defaults to `false`. Default here means actual story names ("Documentation") were not included for sorting and storybook only sorted by the higher level titles ("Components/Data Display/Avatar"). Turning this option `true` means now all story names are included ("Components/Data Display/Avatar/StoryName") up to the lowest level and we can then include them in our nested order array. This includes the documentation pages and we can now sort them however we want.

Also if we don't provide any sorting things are sorted alphabetically which just so happens to put the Documentation page as the first one mostly everywhere. 

## Testing Plan

Go over the PR env and make sure the Documentation pages are listed first under our components and they are navigated to when clicking in a Story.

- Closes https://go.mparticle.com/work/UNI-1214

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -1,17 +1,12 @@
 import type { Preview } from '@storybook/react'
-// import type { IndexEntry } from '@storybook/types'
 
 const preview: Preview = {
   parameters: {
     layout: 'centered',
     options: {
-      // TODO https://mparticle-eng.atlassian.net/browse/UNI-1214
-      // storySort: (a: IndexEntry, b: IndexEntry) => {
-      // console.log('Ordering stories', { a, b })
-      // const order = ['Documentation', 'Cell Types', 'Filters', 'Primary', 'Complex']
-      // return order.indexOf(a[1].name) - order.indexOf(b[1].name)
-      // },
       storySort: {
+        method: 'alphabetical',
+        includeNames: true,
         order: [
           'About',
           ['Introduction', 'Changelog', 'Feedback', 'Component Process'],
@@ -21,7 +16,9 @@ const preview: Preview = {
           [
             'Data Display',
             'Data Entry',
+            ['QueryItem', ['Documentation']],
             'General',
+            ['Button', ['Documentation']],
             'Feedback',
             'Layout',
             'Navigation',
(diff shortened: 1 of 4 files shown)


.storybook/preview.tsx
```diff
@@ -21,7 +16,9 @@ const preview: Preview = {
           [
             'Data Display',
             'Data Entry',
+            ['QueryItem', ['Documentation']],
             'General',
+            ['Button', ['Documentation']],
```
tibuurcio, line 21: Those two were not automatically sorted because they have other stories in it which start with letters before "D", so I manually inserted them here.

src/components/data-display/Badge/Badge.stories.tsx
```diff
@@ -24,7 +24,7 @@ export const DotBadge: Story = {
   },
 }
 
-export const BadgeStatus: Story = {
+export const StatusBadge: Story = {
```
tibuurcio, line 27: Changed because "S" comes after "D" 😬 Also, the other Badge story uses the same format so...

src/components/data-entry/DatePicker/DatePicker.stories.tsx
```diff
@@ -62,11 +62,11 @@ type Story = StoryObj<typeof DatePicker>
 
 export const Primary: Story = {}
 
-export const WithDatePickerWithDisabledYears: Story = {
+export const WithDisabledYears: Story = {
```
tibuurcio, line 65: This is gonna be the Title story in Storybook.

danielsiemens (approved):

PR #805, opened 2026-05-08
fix: form component to preserve static member types in declarations

lukwallace:
#agentic

## Summary / Timeline (i, wallace wrote this)
- attempts to upgrade from 1.65 to 2.3.x
- sees that `Form.useForm` and `Form.useWatch` both evaluate to `any` due to types not resolving correctly from the `.d.ts` file
- sends codex to fix
- results in this PR; pls approve
image

## Testing Plan

---

**PR conventions:**

- Target branch: `main`
- PR title: semantic prefix — `feat:`, `fix:`, or `chore:` (no scopes)
- Branch prefix: matching — `feat/`, `fix/`, or `chore/`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@mparticle/aquarium",
-  "version": "2.3.0",
+  "version": "2.4.0-fix-form-static-member-types.1",
   "lockfileVersion": 3,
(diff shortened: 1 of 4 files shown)


jared-dickman (approved):
lgtm

